PR #2, opened 2015-10-02
Fix errors in mysql script

zhangyy1209:
* Remove an extra comma which will cause error when creating `ADMISSIONS` table
* Remove an extra line which will cause error when creating `DIAGNOSES_ICD` table


tompollard:
many thanks for picking these up @zhangyy1209 - really appreciate your help.

temiwale88:
Hello team, 

How do I edit the mysql script for building mimic. I'm new to using Github. These lines are especially tricky "  FIELDS OPTIONALLY ENCLOSED BY '"' TERMINATED BY ',' ESCAPED BY '/'
	   LINES TERMINATED BY '\n'
	   IGNORE 1 LINES"

PR #3, opened 2015-10-02
Fix date format in mysql script

zhangyy1209:
* Fix date format in loading `CHARTEVENTS_DATA_TABLE.csv` which makes all dates be `0000-00-00 00:00:00`


alistairewj:
Thanks!

PR #16, opened 2015-10-23
Updated references to ICD9 table to diagnoses_icd

joer14:
This query didn't work on my system. I think the references to
ICD9 were the problem because when I replaced them with
diagnoses_icd the query worked fine. Thanks!

### DIFF
--- a/comorbidity/postgres/elixhauser.sql
+++ b/comorbidity/postgres/elixhauser.sql
@@ -1,4 +1,4 @@
-with icd9 as (
+with diagnoses_icd as (
 SELECT icd.subject_id,
        icd.hadm_id,
        icd9_code as code,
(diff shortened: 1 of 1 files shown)


alistairewj:
Could you give more detail on your system?

joer14:
Ahh sorry. I got wrapped up in the moment and thought the error was on your end. 

I forgot to create a schema originally, so needed to replace `mimiciii.` with `public.` . And that was all I needed to do but I thought I needed to replace the icd9 mentions. My bad. 

I hesitated when renaming those values since I thought they were aliases anyway. 

Sorry for the inconvenience.

My new issue is the regenerated hadm_id values, because I'm working with a dataset that was already extracted from mimic ii. But thats a story for another place. 

Thanks for your great work on mimic.

PR #38, opened 2015-12-03
Fixed icustay_details query

tfolkman:
I tried running your icustay_deail sql file with PostgreSQL 9.3.6 and had an error due to the round function's first parameter not being of numeric type. I updated the code to cast the double to numeric and it now runs.


alistairewj:
Good catch!

PR #44, opened 2015-12-19
Added mysql load scripts and Elixhauser calculations

pszolovits:
I'd like to incorporate the scripts I updated for MIMIC-III v 1.3 to load the data into mysql and to compute Elixhauser co-morbidity scores, also in mysql.


tompollard:
Merged! Thanks Pete.

PR #48, opened 2015-12-23
Updating pg copy

ektar:
Hi - thanks for the great work on MIMIC III and using github for sharing!  

I just finished getting the db built in a postgres environment that may be a bit more generalizable - interested in your thoughts on whether this can be integrated.  

Mainly, the current postgres build scripts assume:

* the user has the ability to put data locally on the db server
* the user has the ability to create a superuser account.  

Switching from "COPY" to "\copy", the local psql command, avoids both of these requirements.  Right now the script assumes the data csv files are in the current working directory - I'm not sure how to get \copy to to accept the data directory variable.

My procedure for creating the database was:

    # sudo as postgres, create mimic_rw user and mimic3_v1_3 database
    export PGPASSWORD=<secret>
    export PGUSER=mimic_rw
    export PGDATABASE=mimic3_v1_3
    export PGHOSTADDR=<db_host_addr>
    export SCRIPT_DIR=../mimic_code/buildmimic/postgres/
    psql -f $SCRIPT_DIR/postgres_create_tables.sql
    psql -f $SCRIPT_DIR/postgres_load_data.sql
    psql -f $SCRIPT_DIR/postgres_add_indexes.sql
    # create mimic_ro user with only schema usage and select, this is used by researchers

### DIFF
--- a/buildmimic/postgres/postgres_load_data.sql
+++ b/buildmimic/postgres/postgres_load_data.sql
@@ -48,181 +48,154 @@ SET search_path TO mimiciii;
 --  Load Data for Table ADMISSIONS
 --------------------------------------------------------
 
-\set admissions_csv :mimic_data_dir 'ADMISSIONS.csv'
-COPY ADMISSIONS FROM :'admissions_csv' DELIMITER ',' CSV HEADER;
+\copy ADMISSIONS FROM 'ADMISSIONS.csv' DELIMITER ',' CSV HEADER
 
 --------------------------------------------------------
 --  Load Data for Table CALLOUT
 --------------------------------------------------------
 
-\set callout_csv :mimic_data_dir 'CALLOUT.csv'
-COPY CALLOUT FROM :'callout_csv' DELIMITER ',' CSV HEADER;
+\copy CALLOUT from 'CALLOUT.csv' delimiter ',' csv header
 
 --------------------------------------------------------
 --  Load Data for Table CAREGIVERS
 --------------------------------------------------------
 
-\set caregivers_csv :mimic_data_dir 'CAREGIVERS.csv'
-COPY CAREGIVERS FROM :'caregivers_csv' DELIMITER ',' CSV HEADER;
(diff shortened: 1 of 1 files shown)


tompollard:
Thanks for looking at this. ```\copy``` vs ```copy``` is something that's been bothering me for a while. As you say: 

- the issue with ```\copy``` (which is where we started off) is that it doesn't accept variables. 
- the issue with ```copy``` is that is requires ability to place data on the db server (making it unsuitable for cloud instances like Amazon RDS)

Although I prefer ```\copy``` for the reasons mentioned, it feels like a fairly big limitation not to be able to accept variables. Our internal builds, which run on a Jenkins server, would fail for starters!

If you could update your pull request to use ```\copy``` while dealing with the different directories needed for the Travis and Jenkins build tests, then I'd be happy to merge it. The bit that needs fixing is:

https://github.com/MIT-LCP/mimic-code/blob/master/tests/test_postgres_build.py
``` python
    if os.environ.has_key('USER') and os.environ['USER'] == 'jenkins': 
        # use full dataset
        mimic_data_dir = '/home/mimicadmin/data/mimiciii_1_3/'
    else: 
        mimic_data_dir = curpath+datadir
    call(['psql','-f',fn,'-d',testdbname,'-U',psqluser,'-v','mimic_data_dir='+mimic_data_dir])
```

ektar:
What do you think of this solution - change directory ('\cd') can take a variable, so if we just cd before the \copy calls we get back the advantages of COPY without the disadvantages.

My import routine is now:

    MIMIC_USER=mimic_rw MIMIC_DB=mimic3_v1_3 MIMIC_PASSWORD=secret ./create_mimic_user.sh
    export PGPASSWORD=secret
    export PGUSER=mimic_rw
    export PGDATABASE=mimic3_v1_3
    export PGHOSTADDR=pg_ip_addr
    psql -f postgres_create_tables.sql
    psql -f postgres_load_data.sql -v mimic_data_dir='../../../uncompressed/'
    psql -f postgres_add_indexes.sql

tompollard:
Yep, that should do it. Thanks again :)

PR #18, opened 2015-11-02
Docker postgres

masinoa:
This pull request (full details are in the buildmimic/postgres/README.md file):

* provides necessary Dockerfile to run a container with a Postgres server

* bash scripts on the Docker container automatically run the SQL scripts necessary to load the data from
the csv data files to the Postgres database

* minor changes to the SQL scripts to simplify CSV data loading and reduce error messages by using IF EXISTS clauses

PR #49, opened 2015-12-23
Pg add user

ektar:
Wanted to reduce what's required to be run as superuser (schema is already created in 'create_tables', now just creating user and database), remove dependency (gosu), and make it easier to use database and usernames other than "MIMIC" (e.g. if database has multiple versions of MIMIC3).

Can use with following command:

    MIMIC_USER=mimic_rw MIMIC_DB=mimic3_v1_3 MIMIC_PASSWORD=secret ./create_scmimic_user.sh

### DIFF
--- a/buildmimic/postgres/create_mimic_user.sh
+++ b/buildmimic/postgres/create_mimic_user.sh
@@ -3,7 +3,4 @@
 gosu postgres psql > /dev/null <<- EOSQL
                 CREATE USER MIMIC WITH PASSWORD '$MIMIC_PASSWORD';
                 CREATE DATABASE MIMIC OWNER MIMIC;
-                \c mimic;
-                CREATE SCHEMA MIMICIII;
-		ALTER SCHEMA MIMICIII OWNER TO MIMIC;
 EOSQL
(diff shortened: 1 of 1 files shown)


ektar:
Current pull request is unchanged from current master - looks like master is failing checks?

tompollard:
Many thanks for submitting the pull request - we'll take a look after the Christmas break.

> Current pull request is unchanged from current master - looks like master is failing checks?

Yes, the tests are failing on an unrelated issue (MySQL build scripts).

tompollard:
Hi @ektar, the [create_mimic_user.sh](https://github.com/MIT-LCP/mimic-code/blob/master/buildmimic/postgres/create_mimic_user.sh) script was left over from [a pull request](https://github.com/MIT-LCP/mimic-code/pull/18) that was merged but which introduced a couple of bugs relating to the ```\copy``` vs ```copy``` issue. Thanks for taking a look at it. 

I've merged [several of your commits](https://github.com/MIT-LCP/mimic-code/commits/master), but not all of them because it looks like the later commits are reverting the earlier changes. Please could you take a look at the current status of the files and update the pull request as appropriate?

PR #50, opened 2015-12-24
Completion of pg add user pr

ektar:
Thanks for accepting previous pr so quick!  Those last commits were just to try to figure out why the unit tests were failing (I thought I caused that), but I wasn't sure how to rollback so just created this new branch with the last changes:

- Fixed default db to be lowercase 'mimic' (gave an error when uppercase)
- Fixed bug where new MIMIC_USER variable wasn't being used in user creation


tompollard:
Merged! Thanks Eric.

PR #54, opened 2016-01-06
Update min_surviving_bp.sql

mhw32:
Updating the expire flag query to be consistent with mimiciii notation.


alistairewj:
Just so you know, I changed it from '0' to 0 since the field is numeric. Thanks!

PR #59, opened 2016-01-21
Update icustay_detail hosp_seq generation.

mghassem:
Update icustay_detail generation to correctly identify the hospital sequences, previously was a copy of the icustay_seq.


tompollard:
Pull request is intended to fix https://github.com/MIT-LCP/mimic-code/issues/60

alistairewj:
Thanks for the pull request - I refactored it because I'm pedantic but it looks like it's working now!

PR #61, opened 2016-01-25
don't assume existence of $USER database

gideonite:
When run without an argument `psql` assumes that there is a database of
name $USER. This is not always the case. However, generally there will
be a database called `postgres` which gives a db to login into in order
to create the `mimic` database.


tompollard:
thanks @gideonite

PR #62, opened 2016-01-26
remind users to switch to the MIMIC namespace

gideonite:


tompollard:
thanks again @gideonite

PR #67, opened 2016-01-31
added buildmimic/monetdb

parisni:
Hello,

I added support for monetdb + brief description to install & manage.
I hope it fills your needs

Thanks,

### DIFF
--- a/buildmimic/monetdb/README.md
+++ b/buildmimic/monetdb/README.md
@@ -1,0 +1,46 @@
+# Create MIMIC-III in a local MonetDB database
+
+## Instructions for use
+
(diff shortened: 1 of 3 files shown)


alistairewj:
Thanks Nicolas - I'm going to pull it and run through an install on my machine to see how it does!

alistairewj:
I gave it a go. MonetDB is a bit finicky on Windows but most of it works. I updated the readme as appropriate. I couldn't figure out how to serve a new database and not the demo database though.

Also, the load data script has the path for each load statement as '/path/to/'. Do you know if MonetDB has variables so this only needs to be changed once?

parisni:
It has variable https://www.monetdb.org/Documentation/Manuals/SQLreference/Variables, however I couldn't make it working in the path.  It only works with absolute path.
I keep it in mind, when looking  deeper in documentation.

PR #80, opened 2016-02-19
Update elixhauser-ahrq-v37-with-drg.sql

joel1391:
Added subject_id as a variable for the materialized view. Also added the IF EXISTS clause to DROP MATERIALIZED VIEW to prevent errors for those without an existing view


alistairewj:
That worked perfectly! Thanks!

PR #101, opened 2016-06-30
Updated buildmimic/postgres to import data from .csv.gz instead of .csv

jeblundell:
I thought that rather than decompress all the CSV files first, why not simply decompress it directly into the database? I've not looked into other database formats, but this is certainly supported by postgresql. In its current form here it relies upon a Linux/OS X distribution through its use of the external gzip program, which may be a problem.


alistairewj:
Hey this is great and certainly something we'd like to include. The biggest factor on our mind is compatibility across operating systems. It's pretty hard to stay entirely agnostic, but before we merge the request we'd need to figure out two things:

* Does gzip come with Mac OS X by default? (I believe it does)
* How will this work with Windows? Potentially an if-else statement checking for a local copy of GNU gzip?

jeblundell:
If everything goes via make then we could test if gzip exists and then run a different pair of check/build targets. As in, have a separate gzip SQL too.

jeblundell:
So if we only worry about OS X/Linux, then it's trivially easy: we can check if gzip exists on the path and then run the appropriate SQL. It's actually also possible to massively tidy the code down to not have a bunch of separate ifeq statements. However, Windows apparently doesn't handle for loops very well, so I suspect that the way you're currently doing it is the only way to do it and maintain cross-platform compatability. I've therefore returned the mimic-check and mimic-build targets back to their original form and made separate gz ones. Overall can just run it as "make mimic-gz".

jeblundell:
Also, see above commit: there was a (minor) bug in that "exit1" was specified instead of "exit 1". Either way has the desired effect of bailing though :)

jeblundell:
Re: testing, how would you like to proceed? Presumably it's a bit overkill to try loading the entire dataset even on Jenkins when we're going to do the non-gzip data anyway, so could perhaps do the gzip test data first, drop the tables and then carry on as normal with the rest of testing?

alistairewj:
I agree it makes sense to only test the mini subset using gzip and if it succeeds carry on with the normal testing. If you want to add that it would be great otherwise I'll chat with @tompollard because he's more familiar with the settings on our Jenkins server.

Everything looks great - the only thing I'd consider adding to the makefile is a check to see if some version of gzip exists - maybe a simple check that `gzip.exe -v` exits with status 0. I'm going to run through this on my Windows machine when I get a chance and see double check the caveats of running it. Mainly I want to check if you can just have GNU gzip in the same folder and it will work or whether you need to have it on the path. The former simplifies the install process I think. Going to merge this pull request for now though because I think what I'm talking about is an added feature.

jeblundell:
Am happy to attempt to add it myself, will push a commit when I get back in the country.

PR #103, opened 2016-07-02
Automated building of materialized views

jeblundell:
For your consideration: first pass at using Makefile to automate building of materialized views. Simply pop into, e.g. etc and run "make etc". This particular example shows up the direction I'm thinking of taking it in that this will actually first jump into etc/firstdays and build that first. 

If you like this, then we can create a top-level Makefile that means the user can simply run "make extra" or whatever you want to call it and it'll build the appropriate things with appropriate dependencies built first. Obviously I've only done this for Postgresql, we'll need to do something for MySQL too.


alistairewj:
Hm interesting I didn't consider using make for it. Actually it would be really nice to have the user simply do `make mimic` followed by `make extra` and all the materialized views we have are generated. I've personally never called another Makefile from a Makefile before - is it simple to do?

jeblundell:
Multiple directory makefiles fairly straightforward to do: I think one way that people can follow fairly easily will be having a build target named after the name of the subdirectory that calls "make -C subdirectory buildtarget" (or just omit buildtarget for default; all). I seem to recall that this is the official make way of doing it. See my etc/Makefile for example. 

There's a small catch to using makefiles for materialized views: if you run it then it will always rebuild all the views regardless of if they're already made (as in won't skip the already made ones). For people used to "make clean", that's not ideal. On the other hand, it does mean it deals with depencies nicely and, as you say, you can do it with a single command.

jeblundell:
I'll throw together a first pass of a complete make system and we can go from there. Have essentially got a template now, so shouldn't take me too long.

jeblundell:
So this is what the top-level makefile will look like. I've opted to do $(MAKE) -e -C subdir target (note the addition of -e) so we can change the dbuser/schema etc. for the top-level Makefile only. DATADIR should also be passed down when you run "make mimic datadir=xyz". 

Couple of things:
(1) I'm going to get this up and going with Postgres first, but we should think about how MySQL should be handled. We could do different buildtargets, but it would be nice if we could just set a single command variable, e.g. like PSQL here.
(2) It's not good form to send passwords by environment variable, but it's also very annoying prompting every few minutes for it. Shall we make setting up .pgpass the first thing in build from top-level?
(3) I've been mulling over how to not rebuild the materialized view every time, but at the same time maintain dependency issues. Simplest might be to just test the existence of the view and only create if non-existent. The way to then drop the views from make in the case of a re-build or things going squiffy somehow would be to use the traditional "clean" target. Sound good? So... how best to check from command line that the materialized view exists? Specifically, on Windows. OS X/Linux should be fine.

jeblundell:
Actually, I'm possibly thinking a little too much in Makefile terms here. Issue (3) above is an SQL problem. Surely we should remove all "DROP MATERIALIZED VIEW ..." from the SQL scripts and have a separate "clean" script for when that's required?

jeblundell:
Alrighty, so "make extra" now works as expected: automatically builds all contributed materialized views maintaining dependencies. I've also separated out the DROPs so that "make extra" only builds views that it needs, but you can still do "make clean".

alistairewj:
I'd like to integrate this - but there's been a lot of changes since you initially made this pull request. Any chance you could rebase?

jeblundell:
How's that looking?

alistairewj:
Looks good... One question I had: I noticed at the top level Makefile you have `PSQL=psql "dbname=$(DBNAME) options=--search_path=$(SCHEMA)" --username=$(DBUSER)` - is your plan to use this variable in buildmimic/postgres/Makefile ? And eventually allow for compatibility with MySQL by changing the `PSQL` environment variable in the top level Makefile?

Also trying to think of how to handle inter-folder dependency, e.g. the `severityscores` code requires a large amount of the `etc/firstday` views to be generated. This may be the only folder that has this issue though - perhaps the easiest solution is to just add `@$(MAKE) -e -C etc extra` to the `severityscores: etc` in the top level Makefile. What do you think?
 
Also, we do have the issue with requiring multiple authentications being a bit tedious. This is somewhat alleviated by using peer authentication with PostgreSQL (which is what I use) - but maybe afterwards we should add an issue to create a new feature: using .pgpass for authentication (as you mention above).

jeblundell:
Initial intent was to make it so any command line option changes are propagated to all makefiles, but then it became clear that if we could get all SQL engines working with a command of $(ENGINE) $(OPTIONS) $(FILE) then there's very little work in implementing each engine. I figure worst case it might be a minor fudge of the build rules to be $(PREFIX) $(FILE) $(SUFFIX) and have a blank suffix for postgres, but we can figure all that out when we tackle MySQL. 

Yeah, the inter-directory dependencies aren't so great. Another issue: broadly speaking if you put SQL in severityscores, all of etc is made first, if you do something in etc, you know all of firstday is done. You don't typically have to specify individual SQL file dependencies in Makefile - except in firstday. With firstday, the build order is critical. Would be nice to somehow rearrange firstday to not have to manually specify build targets in the Makefile. I'm thinking of future contributors: ideally we'd have them write 1 SQL file (their view/table/whatever) and add one line to clean.sql.

I might have missed something, but I think the etc dependencies already covered in top-level makefile (line 75/76) with, e.g. severityscores having a dependency on that build target; etc/firstday is built within etc, so should be covered.

Ah yes, had forgotten the authentication issue. I've been using .pgpass too. The alternative is to set an environment variable in top-level Makefile. Can be done with a prompt, but that's a cross-platform nightmare.

Broadly plan is:
* PostgreSQL (this pull request)
* MySQL
* Tidy the dependencies
* Data download (see issue #108)
* Figure out authentication

alistairewj:
Yeah that makes sense. It's actually not too much to refactor the firstday queries to not be interdependent, the only real violator is blood-gas-arterial depending on blood-gas - we could just merge the queries into a single file. 

Anyway this looks good thanks so much for the effort - I'll merge it and start to think of some tests to add to our build. I think we'll definitely lose track of making sure the Makefiles are up to date without some unit tests.

PR #104, opened 2016-07-02
Changed ventdurations from a table to a materialized view

jeblundell:
I've updated ventdurations to be a materialized view rather than a table, in fitting with everything else. Unless there was a reason it was a table instead?

I've not adhered to styling guides purely in the interests of minimal-modification of the code: happy to make it clearer/tidier if that's wanted though. I've run a join between the view and the original table and as far as I can tell they are identical.

### DIFF
--- a/etc/ventilation-durations.sql
+++ b/etc/ventilation-durations.sql
@@ -1,144 +1,137 @@
 -- This query extracts the duration of mechanical ventilation
-drop table ventdurations;
-create table ventdurations as
--- First create a staging table with only vent settings from chart events
-with ventsettings as
-(
-  select
-    icustay_id, charttime
-    -- case statement determining whether it is an instance of mech vent
-    , max(
-      case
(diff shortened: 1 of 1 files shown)


alistairewj:
Yep, I think a materialized view is nicer because it maintains the relationship to the dependent tables. For example if the way GCS is calculated is changed and then the severity score views which depend on it are automatically dropped. Doesn't affect this table as much though since it's derived directly from the raw data, but it's nice to be consistent.

Also I noticed you changed `with` to inline queries - which would make it easier to adapt to MySQL. Did you just do that for style?

Thanks for the improvement!

jeblundell:
Yes. Deliberately for easier MySQL integration *shifty eyes* yep, it was honestly entirely stylistic. Also, not really used with so wasn't sure if doing subqueries would let the query planner do some optimisation. 

Most welcome :)

PR #105, opened 2016-07-07
Added requirements.txt to tests/

jeblundell:
Started playing with the test system: noticed that there's not a standard requirements.txt, so adding it here.


tompollard:
Looks good, thanks @jeblundell. Alistair and I are both travelling at the moment, so it may take us a little longer than usual to get through new issues and pull requests...

jeblundell:
Thanks and no worries: the automated build stuff in particular needs a fair bit of consideration before merging, so wasn't expecting it to be too soon :)

PR #109, opened 2016-07-10
Changed ventfirstday from a table to a materialized view

jeblundell:
I've also only just noticed that ventfirstday is also a table and not a view. Have updated in similar way to #104


tompollard:
Thanks Jim, your changes have been merged. If you are introducing stylistic changes in the future (e.g. WITH clause vs subquery) please could you briefly outline the reason? The original author may have selected a particular style intentionally, so it's good to have justification for overwriting their work. Plus not that I care *shifty eyes* but smaller changes are also quicker to review.

jeblundell:
Thanks very much, will do.

PR #112, opened 2016-07-18
lods.sql

jgieringer:
lods.sql is an implementation of Logistic Organ Dysfunction Score, created by Le Gall J, et al.
It was used to help understand sepsis criteria and thus update its definition given by Singer M, Deutschman CS, Seymour C, et al. "The Third International Consensus Definitions for Sepsis and Septic Shock (Sepsis-3)" 
(http://jama.jamanetwork.com/article.aspx?articleid=2492881)


tompollard:
Thanks for the contribution :) The table builds correctly in Postgres 9.4.5 and at first glance everything looks fine to me. We'll take a slightly closer look in the next few days and will then merge.

tompollard:
This comment is to add Table 5 of the paper which outlines the scoring system. Annoyingly the 1996 JAMA paper ("[The Logistic Organ Dysfunction System: A New Way to Assess Organ Dysfunction in the Intensive Care Unit](http://www.ncbi.nlm.nih.gov/pubmed/8769590)") is still paywalled!:

screen shot 2016-07-20 at 11 45 03

tompollard:
I have listed a few thoughts below after a first pass:

### General

- For clarity, it might help to make the headers in the comments consistent with Table 5 of the paper (i.e. Coagulation -> Hematologic; Respiration -> Pulmonary etc).

### Pulmonary:

- ```platelet_max >= 50.0``` is a condition required for a score of 0, so ```platelet_max >= 50.0 then 1``` is probably intended to be ```platelet_max >= 50.0 then 0```

- Is there a reason why a ```platelet_max < 50.0``` condition isn't present?

### Cardiovascular:

- Shouldn't ```SysBP_Max >= 270``` be assigned a value of 3, not 5?

### Renal:

- ```(UrineOutput >= 500 and Creatinine_Max <= 740)``` should read ```(UrineOutput >= 500 and UrineOutput < 750)```

- ```(Creatinine_Max >= 1.20 and Creatinine_Max <= 1.59)``` should probably be ```(Creatinine_Max >= 1.20 and Creatinine_Max < 1.60)``` for consistency.

### Hepatic:

- the minimum ```PT_Max``` in the labs table is 8 (```select min(PT_Max) from labsfirstday;```) so everyone with a ```PT_Max``` measurement is assigned a score of 1. Is this an issue with the unit?

alistairewj:
Thanks for the contribution. I wrote something similar for another project but didn't add it to this repository, which I probably should have. I'm going to pull yours then update it with my own and we can look at the diffs. Aside from the typos, the biggest issue I found was calculating prothrombin time (PT). It's not an absolute value but rather seconds *above* the standard value. It's also not clear whether the % below standard that they use is a quick value (i.e. `PT / PT_normal`) or the actual standardized value called the international normalized ratio = `( PT / PT_normal )^c` , where c is a reagent specific coefficient. Going to dig into some lab reports later and see if the standard value for PT ever changed.

alistairewj:
So biggest changes are:

* @tompollard all your concerns were addressed
* score components are null if *all* of the necessary variables are null, for example renal is null iff BUN, UO, and creatinine are null
* still haven't determined the reference for prothrombin time

Also, worth noting that the version of LODS used in the paper *The Third International Consensus Definitions for Sepsis and Septic Shock (Sepsis-3)* is a modified version of LODS: it does not use PT, blood urea nitrogen, or urine output.

tompollard:
Thanks @alistairewj. Just adding Table 6 here too, which includes useful additional notes on the scoring system. 

screen shot 2016-07-20 at 15 08 54

jgieringer:
Awesome. Thanks for the checks and corrections!

Issue #114, opened 2016-08-09
Should inr_min or inr_max be used in lods.sql?


jgieringer:
Line 108 selects:
`labs.inr_min`

Line 204 inspects:
`when inr_max is null`

Thus the error output: (I'm unsure why it states line 154, as I am using the most up-to-date version)

> psql:/mnt/mimic/mimic-code/severityscores/lods.sql:234: ERROR:  column "inr_max" does not exist
> LINE 154:       when inr_max is null

alistairewj:
Hey sorry for the slow response on this - got a bit snowed in. The answer is neither should be used. LODS uses Prothrombin Time (PT) in their scoring system - roughly speaking it's how quickly your blood coagulates. PT is quite a difficult lab to analyze because it is absolutely dependent on the laboratory reagents - a PT of 13 seconds could be normal in my lab but abnormal in yours. More relevant to MIMIC - since the laboratory reagents change over the years, a PT of 13 seconds may be abnormal for one patient (admitted in 2002), but the exact same value of 13 would be normal for another patient admitted at another time (say 2005 or so). This is because the lab's reagent change over time, and consequently the reference ranges change over time. In MIMIC, reference ranges have varied between [9.4, 12.5] and [11.3, 13.3].

Now, people are aware of this, and thus the international normalized ratio (INR) was introduced, which normalizes PTs to a standard reference range using a lab-specific normalization constant. Unfortunately LODS does not use INR, either because it predates INR or they didn't use INR where it was developed, I'm not sure.

Back to LODS, the paper states:

> Prothrombin time (seconds or %): If there are several values recorded, assign 1 LOD point if the prothrombin time was ever more than 3 s above standard or less than 25% of standard during the day. 

We need to define "standard". They never really state what standard is - reference ranges are given as [Low Value, High Value], so for example if the reference range is [11, 13] we could interpret "standard" as 13 (the upper value) or 12 (the average value). A similar story can be said for the lower value. And of course, that's on top of the changing reference ranges I mentioned earlier...

To make a long story slightly longer, I've updated the code as follows:

* I have defined standard as 12 seconds. While the reference range changes, it always contained 12, and for most patients this was near the arithmetic mean
* All mentions of INR are removed and replaced with PT

Just wanted you to be aware of all the considerations that have to be made when using this code! Also worth noting that the mLODS (used by Sepsis-3 authors) avoids this issue by not using PT (since they didn't have it in their database). Closing this now - feel free to ask any further questions though.

PR #115, opened 2016-08-09
Update mlods.sql

jgieringer:
Line 193 had an excluded variable, inr_max uncommented-out.


alistairewj:
Thanks - writing a detailed response about this in issue #114

PR #123, opened 2016-09-09
Adult patients and patients with infections

rooster06:
MIMIC-III V1.3
The AdultPatients.sql script generate a table view of subject_id and corresponding hadm_id that belong to patients over the age of 18.

The infections.sql script generates a table view of subject_id, corresponding hadm_id and the start time of the suspected infection.


alistairewj:
Thanks for the contribution. Reviewing your code, I think we shouldn't create temporary tables if they are not necessary, and at the very least should clean them up afterward. Have you seen the icustay_details query here: https://github.com/MIT-LCP/mimic-code/blob/master/demographics/postgres/icustay_detail.sql ? It performs a very similar function to your query except it does not limit the data to adult patients.

Also, the suspected infection query is interesting, but I'd put that in a separate pull request so we can discuss it separately.

rooster06:
Thanks, I will update it and put it in a separate pull request.

zh-zhang1984:
When I run the status code sql, it returns an error message as shown in the following. How can I fix this problem?

ERROR:  column reference "cmo" is ambiguous
LINE 71:   , max(case when rnFirst = 1 then CMO else null end) as CMO...
                                            ^

********** 错误 **********

ERROR: column reference "cmo" is ambiguous
SQL 状态: 42702
字符:2392

rooster06:
hi zh-zhang,

I'm not sure if this question is regarding the code i posted. can you further explain the problem you ran into?

zh-zhang1984:
Yes, it is not your code. I am sorry for the mistake.

PR #126, opened 2016-09-14
Updated buld rule in etc and etc/firstday

jeblundell:
So glossing over how I managed to miss this bit out, here's a little more work on #103 that actually let's it build properly with the new scripts.


alistairewj:
Cheers.

PR #141, opened 2016-10-16
Fixed postgres table check, added ability to use remote pg servers

ektar:


tompollard:
Thanks for the contribution @ektar

PR #142, opened 2016-10-21
Adding in two extra itemids

mghassem:
These correspond to:

 itemid |        label         | dbsource |    linksto     |     category     
--------+----------------------+----------+----------------+------------------
  42273 | vasopressin unit/min | carevue  | inputevents_cv | Free Form Intake
  42802 | VASOPRESSIN  CC/HR.  | carevue  | inputevents_cv | Free Form Intake


alistairewj:
Thanks - this would improve the calculation for two patients but won't work without coalescing amount and rate. I'll add a few fixes then it should work.

PR #144, opened 2016-11-02
Florida Polytechnic Project Codes

alexedurante:
MIT MIMICIII github code repository,

These are the codes we developed during our project at Florida Polytechnic. The codes themselves have to do with generating counts of patients with icd9 codes. Credit for the codes goes to Dr. Gentimis, Kris Kindle, and myself. This was our first time working with PostgreSQL so we welcome any and all advice on improvements.

Thank you for this opportunity to contribute,

-Alex Durante


alistairewj:
Thanks for the contribution. One minor change: instead of having these scripts in their own folder, could you move them to the cookbook/postgres folder? I think it would fit better there.

alistairewj:
That's great! Thanks!

PR #157, opened 2016-12-08
Fix subtracting GU irrigant volume in urine-output-first-day.sql

bshickel:
Current version throws error. This should provide expected functionality for subtracting GU irritant volume.


tompollard:
thanks @bshickel

PR #160, opened 2016-12-14
Fixed type comparison in adenosine-durations.sql

bshickel:
Fixes the following error:

ERROR:  operator does not exist: text > integer
LINE 48:         when vaso_rate > 0 and
                                                  ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.


alistairewj:
Thanks!

PR #177, opened 2017-02-22
Adding precision to los_ and age variables

JohannesNE:
Rounding admit- and dischtime to DATE and then calculating the difference is not very precise. Using EXTRACT(epoch interval) gives actual 4 digit precision. For age the difference is probably negligible, but could be changed for completeness.


alistairewj:
Yep good change, thanks.

PR #189, opened 2017-03-23
Added query folders to their path. Sorted by dependency.

manashty:
Added the query folders. 
It wasn't clear where the queries are located. Two were in the concepts folder and others were in firstday folder.
Also, another dependency was observed and added (ventilation-durations).
Sorted by folder then dependency.


tompollard:
Thanks @manashty, this is a helpful update.

PR #285, opened 2017-08-15
Update sql-intro.md

ZhangAngus:
line 344 miss left bracket befor cast
line 361 also
line 395, 454, 462,506, 519,574, 601  miss semicolon at the end of the code
line 505, 517 temporary table name error, should be icustayorder


alistairewj:
thanks!

PR #286, opened 2017-08-15
Docker friendly

etheleon:
changed mimiciii.table_name to table_name in files within concepts folder


alistairewj:
Awesome thanks for tidying this up.

PR #302, opened 2017-09-05
added analysis of patient mortality

jefftienchow:


alistairewj:
Thanks for the pull request. The main issue is that this code is not reproducible. Looking at your code, you seem to select from a non-raw table (`mimiciii.log_table`), and build your model based on data from that table. To be fully reproducible, you should include code you used to generate the log_table view.

PR #326, opened 2017-11-08
Update directory count in make-concepts.sql

n-s-f:
There are 9 concept subdirectories.

This script echoes something like `\echo 'Directory 1 of 9: durations'` before running the scripts in each one.

For directories 1 through 4, it was in correctly saying 'Directory n of 8:*`


alistairewj:
Thanks!

PR #327, opened 2017-11-08
Fix outdated link in README

n-s-f:
The link in the README to the script that creates the `ventdurations` view is out of data and is leading to a 404.

Thanks
noam


tompollard:
thanks!

PR #389, opened 2018-01-30
generates a row for every hospital admission per patient to describe the counts of their clinical notes.

mso1830:
Title: generates a row for every hospital admission per patient to describe the counts of their clinical notes.
--The statement was created for PostgreSQL and it utilizes a function called crosstab from the tablefunc extension in PostgreSQL. 
--First, we add the extension tablefunc to allow us to use the crosstab function in line 10 
--The statement will show the count and the type of clinical notes written for each patient in each hospital admission.
--The prerequisite for this statement is to create the extension of tablefunc in which we can find crosstab() function.


alistairewj:
Thanks! Seems quite useful. Before merging I modified it a little bit - removed some extra braces and renamed the output table `note_counts` since that seems more intuitive for the table's contents. I also kept the column as `hadm_id` because then it's easier to join to other tables.

alistairewj:
And I probably should have done this before the merge but I just renamed the query note-counts.sql as well for consistency. Thanks again!

mso1830:
thanks for the modifications. it looks better now

PR #393, opened 2018-02-07
Change some variable names and data types of icustay-detail

cx1111:
Small changes. Change string variables (y/n) to boolean. Also lowercase some variable names to be consistent.


tompollard:
thanks @cx1111. anything downstream that this might effect? have you checked views?

PR #410, opened 2018-03-21
Instructions to install MIMIC-III v1.4 to BigQuery

andreysoares:
- Created the folder "bigquery" under "buildmimc"
- Added instructions to install MIMIC3 v1.4 to BigQuery
- Created a sub-folder "schemas"
- Added JSON files with table schemas

### DIFF
--- a/buildmimic/bigquery/README.md
+++ b/buildmimic/bigquery/README.md
@@ -1,0 +1,221 @@
+# Installing MIMIC3 v1.4 to BigQuery
+
+Following are the steps to create the MIMIC-III dataset on BigQuery and load the source files (.csv.gz) downloaded from Physionet.
+
+**IMPORTANT**: Only users with approved Physionet Data Use Agreement (DUA) should be given access to the MIMIC dataset via BigQuery or Cloud Storage. If you don't have access to MIMIC, follow the instructions [here](https://mimic.physionet.org/gettingstarted/access/) to request access.
+
+---
+## STEP 1: Download the MIMIC-III Source files
+
+> NOTE: According to the BigQuery documentation (Last updated March 1, 2018.), "BigQuery can load uncompressed files significantly faster than compressed files because uncompressed files can be read in parallel. Because uncompressed files are larger, using them can lead to bandwidth limitations and higher Google Cloud Storage costs for data staged in Google Cloud Storage prior to being loaded into BigQuery", and "Currently, there is no charge for loading data into BigQuery".
+
+We will proceed using the uncompressed files (.csv). 
+
+### A) Download all MIMIC-III csv.gz files from [Physionet.org](https://mimic.physionet.org/) to your local machine. 
+
+### B) Navigate to the folder you downloaded the csv.gz files and execute the following command to unzip the files.
+
+```
+gunzip *.csv.gz
+```
+
+---
+## STEP 2: Install Google Cloud SDK
+
+### A) Install `google-cloud-sdk`.
+
+```
+export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)"
+
+echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+
+sudo apt-get update && sudo apt-get install google-cloud-sdk
+```
+
+### B) Start Google Cloud. The system should open an Internet browser and ask you to login to Google Cloud Platform and select a default project.
+
+```
+gcloud init
+```
+
+---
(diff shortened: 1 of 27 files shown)


tompollard:
Great, thanks for the contribution @andreysoares. We'll take a look shortly.

paulchurch:
This is great, I was just thinking of adding it.

I've generally used TIMESTAMP instead of DATETIME but both work for mimic3. The distinction would be relevant if multiple timezones were involved.

buildmimic/bigquery/README.md
```diff
@@ -0,0 +1,221 @@
+# Installing MIMIC3 v1.4 to BigQuery
+
+Following are the steps to create the MIMIC-III dataset on BigQuery and load the source files (.csv.gz) downloaded from Physionet.
+
+**IMPORTANT**: Only users with approved Physionet Data Use Agreement (DUA) should be given access to the MIMIC dataset via BigQuery or Cloud Storage. If you don't have access to MIMIC, follow the instructions [here](https://mimic.physionet.org/gettingstarted/access/) to request access.
+
+---
+## STEP 1: Download the MIMIC-III Source files
+
+> NOTE: According to the BigQuery documentation (Last updated March 1, 2018.), "BigQuery can load uncompressed files significantly faster than compressed files because uncompressed files can be read in parallel. Because uncompressed files are larger, using them can lead to bandwidth limitations and higher Google Cloud Storage costs for data staged in Google Cloud Storage prior to being loaded into BigQuery", and "Currently, there is no charge for loading data into BigQuery".
```
paulchurch: I would lean towards uploading the files gzipped because "bq load" on compressed files is still quite fast in practice (a couple minutes) and uploading the uncompressed files can take much more time depending on the network. I switch to uncompressed files only when over the 4GiB hard limit (chartevents is just barely under). Was this a problem in your experience?

buildmimic/bigquery/README.md
```diff
@@ -0,0 +1,221 @@
...
+```
+Waiting on bqjob_r3c23bb4d717cd8a9_000001620e9d5f6d_1 ... (496s) Current status: DONE   
+BigQuery error in load operation: Error processing job
+'sandbox-nlp:bqjob_r3c23bb4d717cd8a9_000001620e9d5f6d_1': Error while reading data, error message: CSV table encountered too many errors, giving up. Rows: 63349; errors: 1. Please look into the error stream for more details.
+Failure details:
+- gs://mimic3_v1_4/CHARTEVENTS.csv: Error while reading data,
+error message: Could not parse 'No' as double for field VALUE
+(position 8) starting at location 3353598526
+FAIL..CHARTEVENTS
+```
```
paulchurch: Consider adding a note at the end to delete the GCS bucket if it's no longer required (gsutil rm gs://mimic3_v1_4/*.csv && gsutil rb gs://mimic3_v1_4). No point paying for storage indefinitely, and safer to clean up sensitive data.

buildmimic/bigquery/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+This code will get the list of files in the bucket, and for each file, it will extract the table name (ex: ADMISSIONS). With the table name, the system executes the BigQuery load command `bq load` to create a table and load the data from the csv file from the bucket using the specific schema from the local folder. The script will need the name of the bucket where the csv files are stored on Cloud Storage, the name of the dataset to create and upload the tables, and the path to the JSON file (schema) on your local machine.
+
+### B) Set the CHMOD to allow the file as executable (ex: 755), and execute the script file.
+
+```
+./upload_MIMIC3_v1_4.sh
+```
+
+During the load with the use of schemas the following error occurred:
```
paulchurch: Since the above script already includes --allow_quoted_newlines, maybe put this documentation up in section A? In this context it might confuse the user into thinking they need to change something.

andreysoares:
Dear Paul,

Thank you for the feedback and suggestions.

I agree that both DATETIME and TIMESTAMP can work with MIMIC data.
> I will include this information to the instructions, so the user can
select either data type.

Regarding the source files, the reason for uploading csv files instead of
gzip, is that we want to keep the csv files for users that would like to
access the csv files via R or Python (if they don't want to use the data
from BigQuery). This would also help to avoid users to create their own
copies (and pay for storage) of the MIMIC raw (csv) files to their Google
Cloud Platform space. We don’t want to keep the gzip file because a user
would need to extract the files before using it and that may involve
additional cost for pre-processing a file.
> I will include instructions to keep the files for those who want access
via R or Python (if needed) and to remove the source files (if not needed).

> I will move the comment about --allow_quoted_newlines to section A.

Thanks,
Andrey

tompollard:
@andreysoares Thanks again for the contribution, which looks good. It sounds like the pull request is awaiting a couple of minor updates, so we'll hold on until merging.

tompollard:
thanks for the contribution!

PR #411, opened 2018-03-22
Change comment re: ventilation-durations.sql location

scottfleming:
mimic-code/etc/ventilation-durations.sql no longer exists. I'm assuming this query needs mimic-code/concepts/durations/ventilation-durations.sql instead?


alistairewj:
Cheers!

PR #453, opened 2018-08-01
code filters icu_length_of_stay at 2 days while comments says 1 day

JackieMium:
in code block 5 `WHERE icu_length_of_stay >= 2` filters the cohort for patients with icu_length_of_stay >= 2 while the comments below says 1 day twice. I've corrected it.


tompollard:
thanks for picking this up

PR #454, opened 2018-08-01
add missing example subject_id

JackieMium:
In the output result above,  the case with `subject_id = 471` has a tranfer service and the ` = 471` seems missing in text.


tompollard:
thanks again @JackieMium

PR #458, opened 2018-08-03
 hadm_id written as icustay_id

JackieMium:
In the above result output `icustay_id` = 291788 has an entry for OMED, 100006 mentioned in the text below is in fact the hadm_id.


alistairewj:
cheers!

PR #474, opened 2018-09-28
Added path after inner joins

steph409:
Hi, 
thanks for the code! Found that some paths were missing. 
Stephi


alistairewj:
thanks!

PR #477, opened 2018-10-01
FIx: add some paths

steph409:
Hi, 
I found some more missing paths as I wanted to create pivoted sofa. 

I also realized that pivot_uo had the wrong path(mimiciii_clinical instead of mimiciii_derived).

Could you maybe create the views on bigquery?

Thank you in advance, 
Stephi :)


tompollard:
@steph409 please could you re-open this pull request when you have finished making changes to the branch?

Issue #484, opened 2018-10-11
Creatinine Stage 3 not exactly as in KDIGO criteria


steph409:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Checked the online documentation: https://mimic.physionet.org/about/mimic/
    * Checked that your issue isn't already addressed: https://github.com/MIT-LCP/mimic-code/issues?utf8=%E2%9C%93&q=

### Description

Hi,

I found a difference in the implementation of the KDIGO criteria (in my understanding of the guideline). 
For stage 3 creatinine, you say in [mimic-kdigo-code](https://github.com/MIT-LCP/mimic-code/blob/master/concepts/organfailure/kdigo-stages-7day.sql), line 60: 

`when HighCreat7day >= (LowCreat7day*3.0) then 3
    when HighCreat7day >= 4 -- note the criteria specify an INCREASE to >=4
      and LowCreat7day <= (3.7)  then 3 -- therefore we check that adm <= 3.7
-- TODO: initiation of RRT`

[kdigo-guideline](https://kdigo.org/wp-content/uploads/2016/10/KDIGO-2012-AKI-Guideline-English.pdf), page 24: increase of 0.5 (so the above should check for <=3.5. 

Is there a medical reason or implementation error? :)

is there a medical reason you implemented the following threshold for KDIGO

JackieMium:
I think you are looking the the wrong place.  Here is a snapshot of the pdf at page 8:

![screenshot_2018-10-14_21-29-13](https://user-images.githubusercontent.com/12944832/46917331-72395880-cff8-11e8-912f-fe3aeaafd5b2.png)

As you can see, the code you pasted is trying to do the 2 items in the red rectangle in this picture.

steph409:
Hi, 
no I only mean the second part of what you highlighted:

> Increase in serum creatinine to >=4.0. 

In the code I cited you can see that the `LowCreat7day` has to be below 3.7. On page 21 of the document I sent, it says that there has to be an increase of at least 0.5, so the code should check for `LowCreat7day` below 3.5. 

It just depends on how you define `increase`

![image](https://user-images.githubusercontent.com/33083326/46917946-37083c80-cf9a-11e8-90ff-d87d181bd1b3.png)

JackieMium:
OK, my bad. I didn't realize you were refering to the 2nd line. 
Yes, I agree with you, I think it should be 4-0.5=3.5.

[...]

PR #486, opened 2018-10-14
Change threshold from 3.7 to 3.5

steph409:
Hi, 

I changed the two lines where the value is not as defined by KDIGO (see Issue #484). 

Bests, 
Stephi


alistairewj:
3.7 was the correct value to use; but the overall logic is being further discussed in #584.

PR #504, opened 2018-11-20
Update sapsii.sql

lauracosgrove:
I noticed when running the query that I encountered an error that the  relation "bloodgasfirstdayarterial" did not exist. This is required in line 115. This was not included in the comment describing the list of necessary materialized views to generate prior to running the query, so I am suggesting its inclusion.


alistairewj:
Oh yep that's definitely an oversight thanks!

PR #514, opened 2019-01-03
Making Dockerfile friendly for auto-builds

haobibo:
Why: Use `cp buildmimic/docker/setup.sh /docker-entrypoint-initdb.d/` instead of `ADD setup.sh /docker-entrypoint-initdb.d/`, because in auto-builds (like dockerhub), the system will use the root folder of this repo as root.

This PR make it easier to make auto builds by refactoring the Dockerfile. Reference case: https://hub.docker.com/r/qpod/postgres-mimic/tags

### DIFF
--- a/buildmimic/docker/Dockerfile
+++ b/buildmimic/docker/Dockerfile
@@ -16,9 +16,8 @@ RUN echo "buildmimic/postgres/" >> .git/info/sparse-checkout
 RUN git pull origin master
 
 # copy the build scripts into a different folder
-RUN cp -r buildmimic /docker-entrypoint-initdb.d/
+RUN cp -r buildmimic /docker-entrypoint-initdb.d/ \
+ && cp buildmimic/docker/setup.sh /docker-entrypoint-initdb.d/
 
 # make a directory for the data
 RUN mkdir /mimic_data
-
-RUN cp buildmimic/docker/setup.sh /docker-entrypoint-initdb.d/
--- a/buildmimic/docker/Dockerfile
+++ b/buildmimic/docker/Dockerfile
@@ -3,19 +3,22 @@ FROM postgres:latest
 # in the docker initialization, we do not build the data
 ENV BUILD_MIMIC 0
 
-RUN apt-get update && apt-get install -y vim git
+RUN apt-get update \
+ && apt-get install -y vim git \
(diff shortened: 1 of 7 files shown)


haobibo:
😃thank you!

xiaosongz:
@haobibo 
Hi Bibo, I am trying to build the mimic docker using your code, and it throws error starting from line 51, it seems have problem recognizing the "1" from `$${PG_VER:0:1}` as an integer, the error looks like:

`/docker-entrypoint-initdb.d/setup.sh: line 51: [: 1{PG_VER:0:1}: integer expression expected
/usr/local/bin/docker-entrypoint.sh: running postgres_create_tables_pg.sql`

And I think this error ultimately triggers the next error, in current version of `buildmimic/Postgres` there no longer exists an file called ` postgres_create_tables_pg.sql` I believe it might be renamed as `postgres_create_tables.sql`.  [buildmimic/postgres](https://github.com/MIT-LCP/mimic-code/tree/master/buildmimic/postgres). I understand since the current `Postgres` docker image is version 12.1, thus even if this file does exist it will still throw more error in the upcoming steps. 

```/docker-entrypoint-initdb.d/setup.sh: line 56: /docker-entrypoint-initdb.d/buildmimic/postgres/postgres_create_tables_pg.sql: No such file or directory```

I am now tapped into the docker and manually running the code after line 51, things runs fine for now. 

Thanks a lot for making this process a lot easier for us!

haobibo:
Thank you @xiaosongz for your feedback!
I will take a look ASAP (it's Chinese new year these days).

xiaosongz:
> Thank you @xiaosongz for your feedback!
> I will take a look ASAP (it's Chinese new year these days).

Hi Bibo,

Happy Chinese new year! No worries, I was able to run the rest scripts manually by tap into the docker container.

I am not very familiar with either docker nor shell script, but I checked that the $PG_VER did resolves as "12.1". Some how the` [ $${PG_VER:0:1} -eq 1 ]; ` seems not correctly resolves as `True`. 

Thank you!

haobibo:
Hi @xiaosongz , I've created a PR trying to fix this bug: https://github.com/MIT-LCP/mimic-code/pull/693

I found the system variable `$PG_MAJOR` provided by docker images `postgres` can be used directly to detect the PG version.

Could you please provide your feedback?

PR #528, opened 2019-02-08
fix missing icd9 code in solid tumour definition

reidhayes:
icd9:173 (Other and unspecified malignant neoplasm of skin) was missing from elixhauser-quan


alistairewj:
Do you have a reference for adding in 173? As far as I'm aware it's not included in the original publication (Table 1 below) or in the code they ship with it: [EnhancedCharlsonICD9CM.txt](https://github.com/MIT-LCP/mimic-code/files/2856518/EnhancedCharlsonICD9CM.txt)
 
quan-icd-codes

reidhayes:
Interesting! Thanks for looking into this.

PR #539, opened 2019-03-12
Update Logic for Central and Arterial Line Durations

andrewbarros:
I was looking at the views for arterial and central line durations and I think we should make a couple of changes:
1) For Metavision Arterial Lines, IABP should be considered arterial access. Also, a sheath without a venous location category is likely arterial -- All of these patients also have arterial blood pressures recorded despite having no other arterial lines. Here is a query I used to confirm this. 
WITH mv_artlines AS (
	SELECT 
		icustay_id
		, COUNT(*) AS totalLines
		, COUNT(location) as NullLocationLines
	FROM
		procedureevents_mv
	WHERE
		itemid = 225789
	GROUP BY icustay_id
	HAVING COUNT(*) > 0 AND COUNT(location) = 0
)
SELECT
	icustay_id
	, COUNT(*)
	, MIN(totalLines)
	, MIN(NullLocationLines)
FROM
	chartevents
INNER JOIN mv_artlines USING (icustay_id)
WHERE
	itemid IN (225309, 225310, 225312) -- MV ArtLine S/D/M BP
GROUP BY icustay_id;

2) IABP should also be considered an arterial line for carevue. Some of the sheaths that are counted as venous access are likely arterial but there's no great way to tease these out.

3) For the central lines in meta vision, I added a couple items. Also, all of these lines should be presumed venous if their location is null (rather than just PICC, port, hickman, and AVA). 

4) For the central lines in carevue, I added several line types that were commonly seen (and should be counted as central lines). The line types are now also sorted by frequency.

### DIFF
--- a/concepts/durations/arterial-line-durations.sql
+++ b/concepts/durations/arterial-line-durations.sql
@@ -6,9 +6,11 @@ with mv as
     pe.icustay_id
   , pe.starttime, pe.endtime
   , case
-      when itemid = 225752
+      when itemid in (225752, 224272)
         then 1
       when pe.locationcategory = 'Invasive Arterial'
+        then 1
+      when itemid = 225789 and pe.locationcategory IS NULL
         then 1
       else 0
     end as arterial_line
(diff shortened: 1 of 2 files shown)


alistairewj:
That all makes sense - thanks so much for contributing these improvements.

PR #549, opened 2019-04-10
Updated README.md in buildmimic/docker to use the correct mount point

JHibbard:
I updated the README.md in buildmimic/docker to use the correct volume mount point for pg_data.  The original capitalization was "/var/lib/PostgreSQL/data", but the "postgres:latest" image writes to "/var/lib/postgresql/data" by default.


tompollard:
thanks!

PR #575, opened 2019-07-12
two tutorials of data visualization

kimheekimi:
I would like to share two jupyter notebook documents for (interactive) data visualization using MIMIC data. I hope you find these useful.


alistairewj:
Thanks for the contribution! A few small things:

1. I note you install some conda packages within the notebook - any chance you can include an environment.yml file so people can prepare the environment beforehand?
2. In the data_viz, toward the middle, you say "Change the path in case your MIMIC dataset is located at different folder". Can you put all user configuration variables at the top? It's easier to adapt and run the notebook.
2. There's a lot of headers with no explanatory text (e.g. "Depict the missing values (2/2)") - I think without an instructor these will be a bit difficult to digest (personally I can only guess what this plot shows). Any chance you can add some more info to the places where you only describe the cell with a header?

kimheekimi:
Hi Alistair Johnson, 
sorry for the delay in response, I was away for holidays and just returned. I will work on things you mentioned on coming Monday and I will be back to you when it is done.

PR #577, opened 2019-07-17
Fix weight unit issue for the lb&oz case

ryanhantruong:
Total patient weight is given by # of lb + # of oz. but the code treated them as separate measurements while they are complementary measurements. Comparison for the same patient between CareVu lb & oz with MetaVision kg weight measurements have verified this. This edit fix the issue and added hadm_id column so the view can be used to aggregate patient stats by hospital stay sessions in addition to ICU stay.

### DIFF
--- a/concepts/demographics/HeightWeightQuery.sql
+++ b/concepts/demographics/HeightWeightQuery.sql
@@ -11,7 +11,7 @@ CREATE MATERIALIZED VIEW heightweight
 AS
 WITH FirstVRawData AS
   (SELECT c.charttime,
-    c.itemid,c.subject_id,c.icustay_id,
+    c.itemid,c.subject_id,c.icustay_id,c.hadm_id,
     CASE
       WHEN c.itemid IN (762, 763, 3723, 3580, 3581, 3582, 226512)
         THEN 'WEIGHT'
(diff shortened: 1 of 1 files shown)


alistairewj:
Thanks! I took the opportunity to clean up the weight-durations code as well, and make this query dependent on it. I've been meaning to refactor the height weight query for a while and your bug report was a good incentive. Just FYI I merged an updated version of your branch where I:

* renamed HeightWeightQuery.sql to heightweight.sql
* added in a birth weight itemid (4183)
* replaced weight processing code in heightweight.sql with the weightdurations table
* explicitly filter out bad heights in this query
* fixed the logic in weightdurations for admission weight removing too many rows
* fixed the logic in weightdurations where echo data wasn't actually imputed

Much cleaner now! And has the weights of neonates with your lb+oz find :)

daiyl:
> Thanks! I took the opportunity to clean up the weight-durations code as well, and make this query dependent on it. I've been meaning to refactor the height weight query for a while and your bug report was a good incentive. Just FYI I merged an updated version of your branch where I:
> 
> * renamed HeightWeightQuery.sql to heightweight.sql ...


Another issue happened when I run the notebook "mimic-code/notebooks/aline/aline.ipynb":
FileNotFoundError: [Errno 2] No such file or directory: '../../concepts/demographics/HeightWeightQuery.sql'

And when I replace the name "HeightWeightQuery.sql" to "heightweight.sql", it is still not work:
No relationship weightdurations 
LINE 76:     FROM weightdurations

alistairewj:
Ah good catch! It needs the weight-durations query to be run first, which is why it still did not work. I just pushed a fix.

daiyl:
> Ah good catch! It needs the weight-durations query to be run first, which is why it still did not work. I just pushed a fix.

It is a pity. It still did not work:
```
No relationship  echodata
LINE 200:   inner join echodata ec
```

dualinsanity007:
Hi. I faced a similar problem when I try to generate the aline dataset using the aline.ipynb.

Below is the snippet of the ipynb.file. I added the postgres-functions.sql and echo_data.sql into the queries. I got this relation error.
image

I tried installing the postgres_make_concepts.sh file on my server before running the ipynb file. Here is a snippet.
image

I'm not sure why is there the relation error, even though the tables can be created.

alistairewj:
Have you tried skipping that cell? If you've already created all the views/tables, there's no need to run that cell.

dualinsanity007:
I modified the .sql files on my backend, by adding an "into table ..." command.
Although this will create a permanent table, it would probably save me from future error.

alistairewj:
OK, good to hear! I've moved most of the code from materialized views to
tables in the repo, so maybe that was a factor.

>

PR #581, opened 2019-07-23
your suggestions have been implemented

kimheekimi:
Hi Alistair Johnson,

I have implemented your suggestions and changed a few things. I hope that's fine with you.

* folder name is changed
* empty data folder is added
* images and resources folders are added
* environment.yml is added
* README is added and prerequisites are explained there
* some narratives are added


alistairewj:
see #587

Issue #556, opened 2019-04-27
problem installing MIMIC3 v1.4 to BigQuery


datavizsolutions:
### Prerequisites

* [x ] Put an X between the brackets on this line if you have done all of the following:
    * Checked the online documentation: https://mimic.physionet.org/about/mimic/
    * Checked that your issue isn't already addressed: https://github.com/MIT-LCP/mimic-code/issues?utf8=%E2%9C%93&q=

### Description

Problem installing MIMIC3 v1.4 to BigQuery

I followed the instructions to installing MIMIC3 v1.4 to BigQuery
(https://github.com/MIT-LCP/mimic-code/tree/master/buildmimic/bigquery)

At step 6 of instructions, I ran the script

The script seems to successfully upload and install on Bigquery except for:

MICROBIOLOGYEVENTS.schema.json

Error message: 
"status": {
    "errorResult": {
      "location": "gs://cst-mimic/MICROBIOLOGYEVENTS.csv.gz", 
      "message": "Error while reading data, error message: CSV table encountered too many errors, giving up. Rows: 1; errors: 1. Please look into the errors[] collection for more details.", 
      "reason": "invalid"
    }, 
    "errors": [
      {
        "location": "gs://cst-mimic/MICROBIOLOGYEVENTS.csv.gz", 
        "message": "Error while reading data, error message: CSV table encountered too many errors, giving up. Rows: 1; errors: 1. Please look into the errors[] collection for more details.", 
        "reason": "invalid"
      }, 
      {
        "location": "gs://cst-mimic/MICROBIOLOGYEVENTS.csv.gz", 
        "message": "Error while reading data, error message: Could not parse '2156-04-13 00:00:00' as date for field CHARTDATE (position 3) starting at location 219", 
        "reason": "invalid"
      }

PATIENTS.schema.json

Error message: 

"errorResult": {
      "location": "gs://cst-mimic/PATIENTS.csv.gz", 
      "message": "Error while reading data, error message: CSV table encountered too many errors, giving up. Rows: 1; errors: 1. Please look into the errors[] collection for more details.", 
      "reason": "invalid"
    }, 
    "errors": [
      {
        "location": "gs://cst-mimic/PATIENTS.csv.gz", 
        "message": "Error while reading data, error message: CSV table encountered too many errors, giving up. Rows: 1; errors: 1. Please look into the errors[] collection for more details.", 
        "reason": "invalid"
      }, 
      {
[...]

PR #588, opened 2019-08-08
Fix BigQuery type errors on load

joshrisley:
I experienced the same `DATE` type error described in https://github.com/MIT-LCP/mimic-code/issues/556, which prevents the load of three tables in BigQuery.

A number of date fields in the CSV files include 00:00:00. Evidently the BigQuery load has become more strict in its `DATE` type parsing, now raising an error on these values.

This fix is consistent with README's description of how `DATETIME` is handled in the schemas.


alistairewj:
cheers!

PR #596, opened 2019-08-19
Updated path for aline_data.csv

JamesSWiggins:
Changes to the directory structure of the AWS aline files necessitated a pathing change.  Now the aline_data.csv file path is relative instead of absolute.


JamesSWiggins:
This has typos.  I will resubmit.

PR #615, opened 2019-09-20
Fix SUBJECT_ID in INPUTEVENTS_CV for MySQL

ibacher:
The current schema defines SUBJECT_ID as a SMALLINT, but it is defined as a MEDIUMINT in all other tables. This simply fixes that oversight.


ibacher:
Withdrawn. Improved PR to follow.

PR #616, opened 2019-09-23
Update MySQL schema to more closely mirror the PostgreSQL schema

ibacher:
This is a major update to the MySQL schema to make it more useful. This converts all `VARCHAR` fields to match the size of those in PostgreSQL and ensured all integer types are consistent across tables (see #615). I've also changed the constraint script to so that it can be easily run whether the foreign keys are already in the database or not, essentially changing `ALTER TABLE DROP FOREIGN KEY ...` to the equivalent of `ALTER TABLE DROP FOREIGN KEY ... IF EXISTS`. Finally, I've converted all `FLOAT` fields to `DECIMAL` fields. In the current schema, some `FLOAT` fields were being truncated or dropped on load.


alistairewj:
Great fixes - thanks so much!

PR #619, opened 2019-09-30
include itemid column in d_items table of cloudformation template

uloma07:
'itemid' column was missing from d_items table in the AWS Cloudformation template script.


alistairewj:
Cheers!

PR #659, opened 2019-12-31
Fix where Postgres versions > 10 were not recognized as supporting new syntax

### LINKED ISSUE #658: Oldest supported version of Postgres?
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked the online documentation: https://mimic.physionet.org/about/mimic/

### Description

What is the oldest version of postgres that is currently supported by MIMIC?

I'm trying to build a copy of mimic in a singularity container and I noticed this line in buildmimic/postgres/Makefile:

    # if psql version is 10, then we can use declarative partitioning
    # this is done in the postgres_create_tables_pg10.sql file
    PSQL_MAJOR_VERSION := $$(psql --version | perl -lne 'print "_pg10" if / 10.\d+/')

- This doesn't correctly identify postgres versions > 10 
- It adds a perl dependency (I think it's the only usage of perl in the codebase)

I'm happy to supply a PR to resolve both these issues but figured it was worth checking that PG9 is still supported. If it's not, I'll supply a PR to drop PG9 support.

ajb5d:
Fixes #658

### DIFF
--- a/buildmimic/postgres/Makefile
+++ b/buildmimic/postgres/Makefile
@@ -29,9 +29,13 @@ endif
 # add ON_ERROR_STOP so scripts stop when they fail
 DBSTRING +=  options=--search_path=$(DBSCHEMA)
 
-# if psql version is 10, then we can use declarative partitioning
+# if psql version is greater than 10, then we can use declarative partitioning
 # this is done in the postgres_create_tables_pg10.sql file
-PSQL_MAJOR_VERSION := $$(psql --version | perl -lne 'print "_pg10" if / 10.\d+/')
+PSQL_SERVER_VERSION := $(shell echo SHOW SERVER_VERSION_NUM | psql "$(DBSTRING)" -At)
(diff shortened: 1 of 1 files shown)


alistairewj:
Thanks - what is `shell` here? Is it some sort of alias for `sh`, `zsh`, or `bash`?

ajb5d:
To be honest, I stole the structure from here:

https://stackoverflow.com/questions/10024279/how-to-use-shell-commands-in-makefile#10024479

My understanding was that it was just the way to call a shell out from GNU Make.

    --ajb

On Jan 2, 2020, at 09:09, Alistair Johnson <notifications@github.com> wrote:

﻿

Thanks - what is shell here? Is it some sort of alias for sh, zsh, or bash?

alistairewj:
Ah right it's a Make thing.. well I tested it on 8.4/9.4/10/11 and it works so cheers!

PR #678, opened 2020-01-27
Add glucose analysis notebooks

lawbaker:
I've added two notebooks in a folder called glucose_analysis. They are for recreating figures and results in a paper we submitted recently with Leo.


lawbaker:
This PR includes notebooks for reproducing a submitted paper (with a later edit for re-submission code)

This is my first PR, so let me know if there are issues.

alistairewj:
As you can see we are not particularly good at reviewing PRs, but in general I've come to the conclusion that your supporting material for a research paper should go in its own repository with a DOI (e.g. via Zenodo).

PR #697, opened 2020-02-26
Adds support to SQLite

tiagofassoni:
Sorry about the delay, but here it is. Added a Python script that creates a SQLite database.

It is made as a Python script because that was the most portable way I can think of, specially given the intended audience probably already has Python and Pandas installed.


tompollard:
I found this useful for building a SQLite file of the demo. Thanks for the contribution! It was merged in https://github.com/MIT-LCP/mimic-code/pull/742

PR #705, opened 2020-03-14
Updating docker entrypoint.sh

fernandoandreotti:
Previous setup would write database to `public` schema. This change guarantees it will go onto `mimiciii` schema which is more comparable with the instructions from the website https://mimic.physionet.org/tutorials/install-mimic-locally-ubuntu/


alistairewj:
Thanks @fernandoandreotti  :)

PR #712, opened 2020-03-28
Fixed UO summation error

AndersMunk:
Kdigo_uo on Bigquery has an error in UrineOutput_6hr, UrineOutput_12hr and UrineOutput_24hr. 
It is supposed to be a sum of the last 6,12,24 hours, but is a summation of the next 6,12 24 hours.

Before:
image

After:
image


alistairewj:
Great catch!

PR #728, opened 2020-04-22
Fix building PostreSQL database in docker fails

emield12:
Adding the indexes to the PostgreSQL database fails as there is a "-" before the SQL command. 
Fix: Removed the "-".


tompollard:
thanks!

PR #734, opened 2020-04-29
Adds executing kdigo-stages.sql into make_concepts.sql

stevesquare:
Adds `kdigo-stages.sql` before kdigo-stages-7day.sql to fix the error below

```
psql:organfailure/kdigo-stages-7day.sql:66: ERROR:  relation "kdigo_stages" does not exist
LINE 12:   INNER JOIN kdigo_stages k
```


alistairewj:
Thank you!

PR #781, opened 2020-07-10
update instructions to use physionet GCS bucket

alistairewj:
* Suggests users to query directly from PhysioNet
* Uses the PhysioNet GCS bucket to load data into BigQuery


tompollard:
good idea

PR #793, opened 2020-08-13
fix psql error: function round(double precision, integer) doesn't exist

JackieMium:
Postgres complains about "ERROR:  function round(double precision, integer) does not exist", I don't know if this is the proper fix but it works. Feel free to correct, I am just PRing this to raise the issue here.


alistairewj:
The repo is in BigQuery syntax now - could you replace the multiple `::numeric` calls with a single `CAST( ... AS NUMERIC)`? That way it works on both SQL flavors.

JackieMium:
Yes of course

PR #809, opened 2020-09-18
Update colloid_bolus.sql

kellyspeth:
1) Edit datatable names so they are compatible with BigQuery; 
2) also change from UNION to "UNION ALL"


alistairewj:
managed to cherry-pick this onto main, thanks!

PR #812, opened 2020-09-30
Blood Glucose Management in the Intensive Care Unit

aldo-arevalo:
Contains the notebooks and queries for the paper **Blood Glucose Management in the Intensive Care Unit: Curation Process and Insights from a Data-driven Approach**. This content has the principal aim to share our findings for future reuse.


alistairewj:
i think this makes sense in its own repository associated with your paper!

PR #814, opened 2020-10-12
Updated link to ventilation_durations.sql

gkovaig:
Updated link to ventilation_durations.sql


gkovaig:
Updated link to ventilation_durations.sql

PR #816, opened 2020-10-15
changed weightdurations to weight_durations in durations/*_dose.sql

strayMat:
The calls to the weight_durations table for doses (epinephrine and norepinephrine) missed a _ . 
It would be better for these calls to be consistent with the table name at creation in make-concepts.sh and postgres_make_concepts.sh.


alistairewj:
Good catch, thanks!

PR #876, opened 2021-03-27
Update adenosine_durations.sql

ZhangAngus:
According to the `DATETIME_DIFF` function definition in file /mimic-code/concepts/postgres-functions.sql, time scale indicators should be wrapped in single quotes.
Otherwise, error will occurs that the field, for example "Hour", does not exist.

### DIFF
--- a/concepts/durations/adenosine_durations.sql
+++ b/concepts/durations/adenosine_durations.sql
@@ -210,7 +210,7 @@ select
   -- generate a sequential integer for convenience
   , ROW_NUMBER() over (partition by icustay_id order by starttime) as vasonum
   , starttime, endtime
-  , DATETIME_DIFF(endtime, starttime, HOUR) AS duration_hours
+  , DATETIME_DIFF(endtime, starttime, 'HOUR') AS duration_hours
   -- add durations
 from
   vasocv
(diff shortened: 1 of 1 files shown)


ZhangAngus:
the same issues are also found in other durations related sql files.

alistairewj:
This is a fundamental incompatibility with the BigQuery syntax (which expects unquoted date parts) and the postgresql syntax (I don't believe it's possible to include a date part data type in a function definition). The regular expression at the top of the bash script does this replacement by quoting the date part: https://github.com/MIT-LCP/mimic-code/blob/f5ffa1407b45c26faa944d3301981f97271aecc0/concepts/postgres_make_concepts.sh#L6

alistairewj:
Also - it's *much* easier to review these changes if they are all in the same PR - and that's a more appropriate use of PRs anyway. You might want to check out some of the guidance on how to do this: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork

ZhangAngus:
Sorry, I only execute several sql files rather than building the whole `concept` directory with `postgres_make_concepts_windows.bat` file. Line 11 shows that you already noticed the single quote issue.
Thank you for your reply.

PR #911, opened 2021-05-12
Add import.sh, a shell script to import MIMIC into SQLite

tmward:
Hi,

Thank you for this fantastic dataset. The level of documentation for all the tables/variables is amazing, and I really appreciate your commitment to open-sourcing all the software.

I noticed that your python to SQLite import script needed pandas, data-chunking (since pandas loads all the data by default into RAM for a single table), and manual editing to run (e.g., changing .csv.gz to .csv) if the user wanted to load the demo dataset.

I therefore made a light, shell-only, script called "import.sh" that only needs SQLite, a POSIX-compliant shell, and gzip.

It follows the naming conventions of the pre-existing import.py, so the output file is "mimic3.db" and tables are named in all lower-case from the file name. For example, the data in "ADMISSIONS.csv" is saved in the "admissions" table.

It works both with the demo dataset (.csv files) and the full dataset (.csv.gz files) with no modification.

RAM and disk utilization is low, as it uses gzip to stream files directly to the sqlite3 binary for loading. It also works quickly: on my
system it took 11 minutes for the full dataset (after a reboot and the caches cleared) and under 5 seconds for the demo dataset.

The shell script was linted with "shellcheck" with no warnings/errors.

I also updated the README to instruct the user on using either import.py or import.sh.

Thanks!

Thomas

### DIFF
--- a/mimic-iii/buildmimic/sqlite/README.md
+++ b/mimic-iii/buildmimic/sqlite/README.md
@@ -10,6 +10,8 @@ into memory. It only needs three things to run:
 2. [SQLite]([https://sqlite.org/index.html)
 3. gzip (which is installed by default on any Linux/BSD/Mac variant)
 
+**Note:** The `import.sh` script will set all data fields to *text*.
+
 `import.py` is a python script. It requires the following to run:
 
 1. Python 3 installed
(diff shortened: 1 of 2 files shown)


alistairewj:
thanks! I rebased it - I also added a note that the shell script loads in the data as `TEXT` fields.

PR #1044, opened 2021-05-25
Add MariaDB configuriation with mimic-iii

jasonjason12570:
add maria DB configuriation


jasonjason12570:
with mimic-iii

tompollard:
Thanks @jasonjason12570. It's been a while since I used MariaDB, but my understanding is that it is very similar to MySQL (or it used to be, early after it was forked from MySQL). 

Are the differences now significant enough that we need to maintain two distinct sets of build scripts (i.e. one set for MySQL and another for MariaDB)? 

I notice that your readme discusses an "XAMPP" tool. Please could you explain what this is, and why it is needed here?

jasonjason12570:
Hi Tom !
Thanks for comments to me to reply the pull request.
First of all I want to say that I'm corporate with the hospital in the part
of datamining.
I find that the researcher of the partners or the other ones who want to
build the mimic database are all ask me if he can build on MariaDB or not.
So I look for any solution to query the mimic iii on that nearly ,so on I
add the configuration to the repository.

I think that MariaDB is friendly for the people who not major in the
computer science.(i.e. the doctor or other field of master who wanna to
learn about it.)
I wanna to explain about the "XAMPP" tool , its popular for the other mayor
who wanna to build a simple database with graphic installing for the
expirement.
XAMPP is an open-source package for PHP development environment containing
MariaDB, PHP, and Perl.
So that's why I add the MairaDB because of the "XAMPP" tools packaged them
together.

In my opinion I think although the MairaDB is similar to MYSQL but the
MairaDB is an Open source database and have some different below.

   1. MariaDB has 12 new storage engines whereas MySQL has lesser storage
   engines.
   2. MariaDB has a larger connection pool supporting up to 200,000+
   connections whereas MySQL has smaller connection pool.
   3. In MariaDB replication is faster whereas in MySQL replication is
   slower.
   4. Comparatively MariaDB is faster than MySQL.

In conclusion, MariaDB comes with numerous inbuilt powerful features and
many usabilities, security and performance improvements.
I think we can add MariaDB configuration to the repository to make more
optional to build.

Thanks for your reply , I just wanna to support the open source project -
"mimic code" !

Sincerely,
Owen

Tom Pollard ***@***.***> 於 2021年5月27日 週四 上午3:54寫道：

PR #1047, opened 2021-05-25
Duration of Octreotide Intake

loberste:
In the past I have worked on extracting durations of chemotherapeutic administration (Octreotide infusion) out of MIMIC-III. Therefore, I was able to summarize certain treatment durations using MySQL.


briangow:
Thanks for the contribution @loberste !

PR #1048, opened 2021-05-26
Add instructions/shell script to import MIMIC-IV into DuckDB

tmward:
Hello,

Thank you for the great dataset. I've moved most my analyses for large data over to DuckDB rather than SQLite, as it's a performant OLAP database that is optimized for analytical queries, while still coming with the benefits of SQLite (single file, serverless model).

I wanted to share the code and README I made so that others could benefit and easily use DuckDB as well.

Full details are in the README, but summarizing, loading MIMIC-IV only requires DuckDB CLI installed (a single binary) and then a posix-compliant shell. Then you download the data and run import_duckdb.sh, which takes care of the rest.

The shell script passes with no warnings/errors from ShellCheck. It also has a built in -h/--help feature and does some basic error checking (e.g., it won't clobber a database that already exists). It will run, unmodified, with .csv or .csv.gz data files. I lifted the create.sql code from the postgres directory to do the correct initialization of the schema/tables, only needing to do minor modifications (remove the precision component to timestamp column declarations and remove two not null constraints).

### DIFF
--- a/mimic-iv/buildmimic/duckdb/README.md
+++ b/mimic-iv/buildmimic/duckdb/README.md
@@ -1,0 +1,111 @@
+# DuckDB
+
+The script in this folder creates the schema for MIMIC-IV and
+loads the data into the appropriate tables for
+[DuckDB](https://duckdb.org/).
+DuckDB, like SQLite, is serverless and
+stores all information in a single file.
+Unlike SQLite, an OLTP database,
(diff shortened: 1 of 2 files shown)


alistairewj:
very interesting! thanks for the comprehensive PR (sorry it took a while to get around to it).

PR #1064, opened 2021-06-10
Remove request for users to add a MIMIC-III, MIMIC-IV, or MIMIC-CXR label when creating an issue

briangow:
Users need to have triage or write access to able to add a label.  We will need to manually add labels to incoming issues internally until an automated method can be implemented.


tompollard:
Oh well, would have been nice to have this! I think it's probably still worth asking people to be clear about which dataset they are referring to.

Issue #917, opened 2021-05-01
NULL aki_stage in the kdigo_stages table


Qinyu-Allen-Zhao:
Hi,

I'm working on MIMIC-IV on the Google BigQuery platform. In the kdigo_stages table, I found there are many records with numbers in the aki_stage_creat column and NULL entries in the aki_stage column.

```
SELECT * FROM `physionet-data.mimic_derived.kdigo_stages` 

WHERE aki_stage_creat > 0

AND aki_stage IS NULL

LIMIT 20
```

![image](https://user-images.githubusercontent.com/71951741/116770343-b46efc00-aa75-11eb-9672-8db5b5376589.png)

I suppose aki_stage is the combination of creatinine and urine output criteria. It should be NULL if there are entries in aki_stage_creat.

SichengH:
Here is the script for generating derived tables. https://github.com/MIT-LCP/mimic-code/tree/main/mimic-iv/concepts/organfailure

NULL in the  "aki_stage" means there is no record.

Please let us know if you have further questions.

Qinyu-Zhao0522:
Hi!

I'm not sure about that. Could you check the results of my query?
These patients have serum creatinine records,  and they were diagnosed as AKI according to the creatinine criteria. However, they had NULL in the AKI_stage column. Patients had AKI_stage only when they both had creatinine and urine output.
Is there anything wrong?

tompollard:
@Qinyu-Zhao0522 From the [kdigo stages query](https://github.com/MIT-LCP/mimic-code/blob/main/mimic-iv/concepts/organfailure/kdigo_stages.sql), it looks like `aki_stage` is the max of (aki_stage_creat, uo.aki_stage_uo)

https://github.com/MIT-LCP/mimic-code/blob/0edce4cea6e3b6ffc4062b64f5b4a32c376b8918/mimic-iv/concepts/organfailure/kdigo_stages.sql#L81-L82

It seems that `GREATEST(cr.aki_stage_creat, uo.aki_stage_uo)` returns null if either of the values is null. I'd need to investigate, but I think this is a bug. 

In cases where one value is null, I assume we would want `aki_stage` to return max of the alternative value. e.g. `GREATEST(1, null)` should probably return 1 not null.

tompollard:
[...]

PR #1084, opened 2021-06-30
Fix kdigo_stages.sql in BigQuery for cases where one of the inputs to GREATEST is null. Ref #917 

tompollard:
As noted in #917, Postgres and BigQuery handle `greatest(null,x)` differently. BigQuery returns Null, Postgres returns x. 

It looks like this is addressed in most queries with `COALESCE` but [mimic-iv/concepts/organfailure/kdigo_stages.sql](https://github.com/MIT-LCP/mimic-code/blob/0edce4cea6e3b6ffc4062b64f5b4a32c376b8918/mimic-iv/concepts/organfailure/kdigo_stages.sql#L81-L82) seems to have been missed.

This pull request fixes `GREATEST(cr.aki_stage_creat, uo.aki_stage_uo)` for BigQuery by  adding the COALESCE.

### DIFF
--- a/mimic-iv/concepts/organfailure/kdigo_stages.sql
+++ b/mimic-iv/concepts/organfailure/kdigo_stages.sql
@@ -79,7 +79,10 @@ select
   , uo.uo_rt_24hr
   , uo.aki_stage_uo
   -- Classify AKI using both creatinine/urine output criteria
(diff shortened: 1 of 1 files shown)


briangow:
Looks good!

PR #1091, opened 2021-07-15
Update some column names used in first_day_lab

kiendang:
03da086 changed some column names in `blood_differential` but they haven't been updated in `first_day_lab`.


alistairewj:
Thanks!

PR #1092, opened 2021-07-15
Change interval x to interval 'x'

kiendang:
for compatibility with postgres


alistairewj:
Thanks! This threw me off for a while, turns out BigQuery is happy with `INTERVAL '4' HOUR`, but will throw an error if you try to use it as a date part e.g. `DATETIME(2020, 1, 1, 0, 0, 0), 'YEAR')` is invalid (but `YEAR` is ok).

PR #1111, opened 2021-08-11
Update directories for MIMIC-IV

briangow:
For consistency with our documentation we should refer to the MIMIC-IV modules by including the -iv- in their name.  Rename mimic-cxr to mimic-iv-cxr. Add mimic-iv-ed and mimic-iv-waveforms directories.


tompollard:
Sounds good to me! Did you intend to add the following files?

- mimic-iv-cxr/buildmimic/postgres/create.sql 
- mimic-iv-cxr/buildmimic/postgres/load.sql

briangow:
@tompollard  I did not intend to include those, thanks for catching that!

PR #1115, opened 2021-08-16
update dockerfile for new git repo layout

ajb5d:
The previous MIMIC-III dockerfile fails for two reasons:

- The default branch is called main (previously master)
- The build scripts are now in a subdirectory of mimic-iii (previous /)

This PR fixes both of these issues.


ajb5d:
The previous MIMIC-III dockerfile fails for two reasons:
- The default branch is called main (previously master)
- The build scripts are now in a subdirectory of mimic-iii (previous /)

tompollard:
looks good, thanks @ajb5d

PR #1120, opened 2021-08-24
PostgreSQL concepts

alistairewj:
Decided to make this easier for everyone:

1. The shell script now generates the postgres SQL files and a `postgres-make-concepts.sql` file in the postgres subfolder
2. The postgres subfolder is now committed to the GitHub repo, noting that the scripts are autogenerated
3. Some READMEs to explain the setup

I sort of tested it .. but it could use another look.


alistairewj:
There were various issues but I *think* they are all fixed so the postgres-make-concepts should run end to end with:

```sql
-- NOTE: many scripts *require* you to use mimic_derived as the schema for outputting concepts
-- change the search path at your peril!
set search_path to mimic_derived, mimic_core, mimic_hosp, mimic_icu, mimic_ed;
\i postgres-functions.sql -- only needs to be run once for the mimic_derived schema
\i postgres-make-concepts.sql
```

JackieMium:
Confirmed!  All 58/58 scripts now worked and generated 58 concept tables without any error.

alistairewj:
awesome, thanks! I'll close the other PR.

kiendang:
This is awesome. Thanks a lot!

PR #1122, opened 2021-08-30
folders to generate dependencies first [postgres_concepts branch]

JackieMium:
I've reordered folders in the second section (2nd loop) in `convert_bigquery_to_postgres.sh` to process scripts in `measurement` and `firstday` folders first and then `score` scripts that depends on them. Now 55/58 tables can be successfully generated.  `score/meld`, `sepsis/suspicion_of_infection` and `sepsis/sepsis3` are not automatically generated.

- `score/meld` can be manualy generated by running `\i postgres/score/meld.sql` in psql. 
- `sepsis/suspicion_of_infection` failed due to `DATETIME_TRUNC` function not found, which meant `sepsis/sepsis3`would have to wait too.  I tried to write `DATETIME_TRUNC` function but then "type 'DATETIME' not found" occurred. This can be solved by replace `DATETIME` to `TIMESTAMP(3)` but I decided to manually edit/process `sepsis/suspicion_of_infection` TWICE is too much of a hack and not worth it. 

 
I also think all except `postgres/postgres-functions.sql` and  `postgres/README.md` in `postgres` folder should be set to `untracked` by git because the sql scripts are automatically generated by `convert_bigquery_to_postgres.sh`. Perhaps the best solution is to put `postgres/postgres-functions.sql` and  `postgres/README.md` just in `concepts` folder and ignore the whole `postgres` folder  for good?
PS: I rebase the `postgres-concepts` branch and seemed to mess the commit log a bit, hope it's fine.


alistairewj:
I adopted these changes into #1120 maybe try with that branch?

PR #1145, opened 2021-09-24
Added mimic_derived schema creation logic for Postgres

tchainzzz:
This commit adds logic for creating the `mimic_derived` schema for building MIMIC in Postgres. This step is necessary to extract concepts using `postgres-make-concepts.sql` to `mimic_derived`; currently, the concept tables are erroneously created in `mimic_core`.


briangow:
Thanks @tchainzzz , this looks good!

PR #1177, opened 2021-10-23
Update postgres_make_concepts_windows.bat

ruiruiandliangliang:


alistairewj:
Thanks - these appear to be local settings so I will close this and leave the more common defaults.

Issue #1179, opened 2021-10-28
Calculating Sofa concept: Several provided scripts can't find several data. MIMIC III


Philippmuench92:
### Prerequisites

* [X] 
    * Checked the online documentation: https://mimic.mit.edu/
    * Checked that your issue isn't already addressed: https://github.com/MIT-LCP/mimic-code/issues?utf8=%E2%9C%93&q=
    * Also tried so resolve the issue following https://sql.info/d/bigquery-error-not-found-dataset-xyz-was-not-found-in-location-us and https://stackoverflow.com/questions/51395256/error-not-found-dataset-my-project-namedomain-public-was-not-found-in-locatio

### Description

Im trying to calculate the SOFA score in MIMIC 3 using the provided concepts from github.
To do so you have to run the 9 subordinate queries from which the following 6 worked without any issues:
-pivoted_uo
-pivoted_lab
-pivoted_gcs
-ventilation_duration
-dopamine_dose
-dobutamine_dose

1. Issue: Running pivoted_bg BigQuery returns me: 

"Not found: Dataset physionet-data-330409:physionet-data was not found in location US at [148:1]"

pointing to line 148. But looking at the directory the table is there as you can see here.

![grafik](https://user-images.githubusercontent.com/92937536/139230964-75693cf9-efbf-4d30-9f0f-a6139b29dd81.png)

2. Issue: Running norepinephrine_dose BigQuery returns me: 

"Not found: Table physionet-data:mimiciii_clinical.weight_durations was not found in location US"

Checking the directory manually the table was indeed not there as you can see here:

![grafik](https://user-images.githubusercontent.com/92937536/139232547-a590c191-1b75-49c3-8c26-d79515ceb513.png)

To resolve this I also tried to run the weight_durations query before and save it into the directory but the system said I dont have permission to do so.

![grafik](https://user-images.githubusercontent.com/92937536/139233063-715d051a-7b94-4ec5-a251-c82688056c87.png)

3. Issue: Running epinephrine_dose BigQuery returns me the same error as for Issue no. 2

"Not found: Table physionet-data:mimiciii_clinical.weight_durations was not found in location US"

This scripts tells me it needs the weight_first_day table which requires echodata view, generated by concepts/echo-data.sql.

Unfortunately trying this also did not help.

I'm really confused since some scripts work just fine and others do not.

My workspace/explorer looks like this:

[...]

PR #1183, opened 2021-11-01
Misc query fixes to MIMIC-III

alistairewj:
Issues identified in #1179

- incorrect table references
- updated make-concepts.sh to generate some more of these concepts


tompollard:
thanks alistair

PR #1191, opened 2021-11-15
Heart failure concept

pedrogemal:
Includes SQL query to fetch ntprobnb lab results and create a concept… named as heart failure.


alistairewj:
What do you think about adding this to the cardiac marker query? https://github.com/MIT-LCP/mimic-code/blob/main/mimic-iv/concepts/measurement/cardiac_marker.sql

pedrogemal:
Hello @alistairewj thank you for the feedback. I had this doubt indeed, since BNP does relate with heart, but is not a myocardial infarct biomarker (as the others from this table). However, thinking about the "concepts", it may be a better idea to join them. Will do the fix and re-submit. Thank you.

alistairewj:
OK - sounds great thanks!

PR #1195, opened 2021-11-29
Allow extra options to be passed to psql calls with MIMIC-III

juliangilbey:
Hi!

The `mimic-iii/buildmimic/postgres/` scripts allow the database to be owned by any user, but the script `mimic-iii/concepts/postgres_make_concepts.sh` does not currently support this. This patch gives the `postgres_make_concepts.sh` the ability to do this, without changing its default behaviour.

Note: the patch looks like it's rewritten every single line of this script, but actually the changes are very modest: the key change is to use a connection string rather than command-line options, which requires that `${CONNSTR}` is double-quoted everywhere.


alistairewj:
Seems reasonable - thanks! Probably worth adding the to the MIMIC-IV version later.

PR #1204, opened 2021-12-09
Include BNP in cardiac markers concept

pedrogemal:
Includes SQL query to fetch ntprobnb lab results in cardiac markers concept. 
Also, removed reference to troponin I, since it is not registered in mimic-iv.


alistairewj:
Wonderful thank you!

Issue #938, opened 2021-05-14
Incorrect LOINC codes in d_labitems


Mauvila:
Is there a plan to correct the possibly incorrect LOINC codes in d_labitems (by either replacing the incorrect LOINC codes or creating a new column in the table)?  The itemids of MIMIC-IV appear to be the same as those from MIMIC-III, and (by random sampling) most of the LOINC codes haven't changed.  

I first want to say that correct LOINC codes are important in this project.  While a hospital system's EMR probably doesn't care about LOINC codes much because it doesn't really affect the EMR's internal operation, the LOINC codes become more important when they are used for database analysis--they allow researchers to use lab based algorithms on multiple databases.  

For MIMIC-III,  I looked at the separate NLM LOINC code determinations (which were provided separately from the MIMIC-III database), and then ran the LOINC codes through my own validator.  Overall, I detected that about 75 of the rows (out of 753 rows total) in d_labitems had likely incorrect LOINC codes.  Of the 75 likely incorrect LOINC codes, the NLM specified codes was correct in about 62 of them (about 80%).  Most of these are instances where the EMR LOINC codes used the wrong LOINC property.  For example, itemid of 50960 has a LOINC code of 2601-3.  This is the LOINC code for serum magnesium in units of moles/volume (property=substance concentration).   The correct code should be 19123-9, which is serum magnesium in units of mass/volume (property=mass concentration).  This is compatible with the only reported unit in the labevents for this (mg/dL).  A few of the incorrect codes are WAY off for other reasons.  A lot of the other likely-incorrect codes were not as clear (based on the MIMIC-III data available).  

I think it would be useful to create a new column in d_labitems that represents the revised LOINC code.  This way, the original LOINC code would still be preserved (useful in the cases where the most correct LOINC code is unclear).

tompollard:
Hi all, @a-chahin is a clinician and informatician who has been working with us on mapping MIMIC to common terminologies (partly for an OMOP project with OHDSI: https://github.com/OHDSI/MIMIC). He also picked some major issues with LOINC mappings and we have been trying to decide on the best way of integrating everyone's efforts.

After discussion with @alistairewj, we agreed it would make sense to drop the LOINC column from the MIMIC data and include the LOINC mapping as a derived table in this repo. This allows us to continue to improve the mapping over time, and avoids the suggestion that the mappings are ground truth.

[...]

PR #1245, opened 2022-02-02
Add initial mapping between itemid and loinc. ref #938.

a-chahin:
As discussed in issue #938, many of the LOINC mappings in the current version of MIMIC IV are not accurate. @Mauvila privately shared a spreadsheet with a number of improvements for the mappings. 

In the next version of MIMIC IV, we plan to remove the `loinc_code` column from `d_labitems` table. Instead the mappings will be provided as a table in the MIMIC code repository. This will allow the mapping to be improved over time.

This pull request adds an initial CSV file with itemid to loinc mapping codes that I think are correct. Using RELMA and clinical knowledge, I manually mapped and verified the LOINC codes. The information I used to make the mapping is `label`, `fluid`, `category`, `valueuom` from `d_labitems` and `labevents` tables.

I have not populated the rows with the mappings provided by @Mauvila, so that he is able to add his work in a separate pull request. I did not want to commit his work without his asking for his permission.


tompollard:
Thanks @a-chahin! @alistairewj, let us know if you have thoughts about the file name, location, structure, etc.  All of the LOINCs listed here are ones where there is disagreement with the spreadsheet that we received. In most or all of the cases, I believe the issue relates to the fluid type. @xborrat is now doing a review of the mappings.

alistairewj:
Looks good - I only have cosmetic thoughts

- why not under concepts? e.g. `concepts/mapping/` or `concepts/concept_map/` or something.
- I would include `itemid` in the filename e.g. `lab_itemid_to_loinc.csv`.
- Somewhere we should specify which version of LOINC we map to. Perhaps even another column, since future versions of LOINC may change the code meaning.

xborrat:
Good job @a-chahin! I have checked one by one all the codes in the `loinc` column and found only 2 minor points to comment:

1/ In `itemid` 50937 which corresponds to "Antibody against hepatitis A virus" the units in the `valueuom` column appear N/A|Pos/Neg unlike the other  yes/no test where only N/A appears. 
2/ `itemid` 51094  which represents the urine pH it would be better to use 2756-5 which corresponds the generic concept of urinary pH than 5803-2 which is too specific concept because is the urine pH test strip.

a-chahin:
> Looks good - I only have cosmetic thoughts
> 
> * why not under concepts? e.g. `concepts/mapping/` or `concepts/concept_map/` or something.
> * I would include `itemid` in the filename e.g. `lab_itemid_to_loinc.csv`.
> * Somewhere we should specify which version of LOINC we map to. Perhaps even another column, since future versions of LOINC may change the code meaning.

@alistairewj Thank you for the feedback! I'll start implementing the changes.

a-chahin:
> Good job @a-chahin! I have checked one by one all the codes in the `loinc` column and found only 2 minor points to comment:
> 
> 1/ In `itemid` 50937 which corresponds to "Antibody against hepatitis A virus" the units in the `valueuom` column appear N/A|Pos/Neg unlike the other yes/no test where only N/A appears. 2/ `itemid` 51094 which represents the urine pH it would be better to use 2756-5 which corresponds the generic concept of urinary pH than 5803-2 which is too specific concept because is the urine pH test strip.

@xborrat Thank you for your feedback!

1/ Regarding the other yes/no tests that have N/A in the `valueuom` column, I checked the `value` column and found out that the values are NEG/POS. For example, `itemid`  50940 "Hepatitis B Surface Antibody" has a N/A in the `valueuom` column. When I ran a query for DISTINCT `value` I got "7.9/NEG/POS"
2/ Excellent point! I think you are right. 2756-5 is better since it does not specify the method. I selected 5803-2 because there is another `itemid` 51491 for Urine pH (but with a category of "Hematology" which is weird) that was mapped 5803-2, and 5803-2 has a higher rank on RELMA than 2756-5 so I assumed this is the most common method to test the pH of urine.

alistairewj:
@a-chahin worth looking at #938 for some additional discussion

a-chahin:
> @a-chahin worth looking at #938 for some additional discussion

Thank you, @alistairewj ! I think it is helpful to check with BIDMC regarding the LOINC mappings since Health Information Exchange (HIE) is one of the requirements of meaningful use stage 3.

tompollard:
Thanks @a-chahin, this is a great start. We can follow up with additional LOINCs and updates to the structure as needed.

PR #1268, opened 2022-03-03
instructions to load dataset to Amazon Athena

yuvallb:


mimic-iv/buildmimic/athena/schema.sql
```diff
@@ -0,0 +1,779 @@
...
+
+
+CREATE EXTERNAL TABLE `mimiciv_csv`.`chartevents`(
+  `subject_id` bigint, 
+  `hadm_id` bigint, 
+  `stay_id` bigint, 
+  `charttime` string, 
+  `storetime` string, 
+  `itemid` bigint, 
+  `value` double,
```
Nintorac, line 92: should be `VARCHAR(200)`

Nintorac (commented):
Hey, thanks for writing this! Saved me some time :D 

I didn't verify the rest of the schemas yet but these came up as issue for us. When I have some time I will verify the rest and update if necessary.

yuvallb:
Thanks [Nintorac](https://github.com/Nintorac)! 
Let me know if there are other issues, and also if all went well.

Nintorac:
There ended up being issues on most of the tables, I pretty much just set every varchar as a string so there might be some more improvements to make but here is the current schema I've come up with


[schema.tar.gz](https://github.com/MIT-LCP/mimic-code/files/9915026/schema.tar.gz)

alistairewj:
Would be happy to merge this - my initial thoughts are:

1. Not familiar with loading data into Athena, but my hunch is you need to specify that the fields are quoted `"`
    * Could be that you need to specify quotes in the info here: https://github.com/MIT-LCP/mimic-code/blob/3c910009d54a20a2439ae2b89e3556116368a2dc/mimic-iv/buildmimic/athena/schema.sql#L18-L19
    * Or it could be that this is misspecified, since the text columns are quoted: https://github.com/MIT-LCP/mimic-code/blob/3c910009d54a20a2439ae2b89e3556116368a2dc/mimic-iv/buildmimic/athena/schema.sql#L26-L33
3. Needs an update for v2.2. The tables in `core` have been moved to `hosp`, and there are new columns.
4. Not sure what the distinction is between the `mimiciv_parquet` and `mimiciv_csv` schemas. In general it would be great to keep the `mimiciv_hosp`, `mimiciv_icu` schema names as that makes concept code much easier to transfer.

PR #1304, opened 2022-05-15
split cd to core subdir to new line

alistairewj:
this avoids issues when mimic_data_dir is not defined, but data is available in the current folder


briangow:
@alistairewj makes sense to me, thanks!

PR #1310, opened 2022-05-27
Update loinc table

a-chahin:
This pull request updates the lab results mapping to LOINC table. Compared to the old table, new columns were added as well as new mappings. The table was reviewed by @xborrat.


danamouk:
thanks @a-chahin and @xborrat , this looks good. @tompollard, and @alistairewj any thoughts on the current file mapping name/columns headers? And anything we should be aware of given the recent modifications to the MIMIC data structure? Further descriptions of the mapping could be found under the Common Vocabulary heading on this page: https://github.com/MIT-LCP/northwestern_federated.

tompollard:
Thanks again @a-chahin and @xborrat!

PR #1339, opened 2022-07-13
Update MIMIC-IV-ED build scripts to v2.0

alistairewj:
Account for the following changes (from the [PhysioNet MIMIC-IV-ED v2.0 changelog](https://physionet.org/content/mimic-iv-ed/2.0/#release-notes)):

> Added additional columns to the edstays table.
> * gender
> * race
> * arrival_transport
> * disposition
>
> The pain column of the triage table is now free-text, and now includes free-text entries which are not valid numbers. This is more consistent with the pain column in the vitalsign table, which was already free-text.


alistairewj:
Opened a new PR for this: #1340.

PR #1363, opened 2022-08-11
mimic-iv/concepts: fix `postgres-make-concepts` and minor updates

schu:
Please see the individual commits for context.

### DIFF
--- a/mimic-iv/concepts/convert_bigquery_to_postgres.sh
+++ b/mimic-iv/concepts/convert_bigquery_to_postgres.sh
@@ -64,10 +64,19 @@ echo " done!"
 # (2) output to the postgres subfolder
 # (3) add a line to the postgres-make-concepts.sql script to generate this table
 
-# order of the folders is important for a few tables here:
-# * scores (sofa et al) depend on labs, icustay_hourly
-# * sepsis depends on score (sofa.sql in particular)
-# * organfailure depends on measurement and firstday
+# we control the order by skipping tables listed in the below var
+DIR_AND_TABLES_TO_SKIP='demographics.icustay_times demographics.weight_durations measurement.urine_output organfailure.kdigo_uo organfailure.kdigo_stages firstday.first_day_sofa sepsis.sepsis3 medication.vasoactive_agent medication.norepinephrine_equivalent_dose'
+
+# create an array to store tables for which the order of generation matters
+# i.e. these tables cannot be generated in alphabetical order, as done in the later loop
+TABLES_TO_SKIP=()
+for dir_and_table in $DIR_AND_TABLES_TO_SKIP;
+do
+  tbl=`echo ${dir_and_table} | cut -d. -f2`
+  TABLES_TO_SKIP+=($tbl)
+done
+
+echo $TABLES_TO_SKIP
 # the order *only* matters during the conversion step because our loop is
 # inserting table build commands into the postgres-make-concepts.sql file
 for d in demographics measurement comorbidity medication treatment firstday organfailure score sepsis;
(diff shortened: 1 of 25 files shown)


mimic-iv/concepts/convert_bigquery_to_postgres.sh
```diff
@@ -4,12 +4,12 @@
...
 export REGEX_SCHEMA='s/`physionet-data.(mimiciv_hosp|mimiciv_icu|mimiciv_derived).([A-Za-z0-9_-]+)`/\1.\2/g'
 # Note that these queries are very senstive to changes, e.g. adding whitespaces after comma can already change the behavior.
-export REGEX_DATETIME_DIFF="s/DATETIME_DIFF\(([^,]+), ?([^,]+), ?(DAY|MINUTE|SECOND|HOUR|YEAR)\)/DATETIME_DIFF(\1, \2, '\3')/g"
+export REGEX_DATETIME_DIFF="s/DATETIME_DIFF\(([^,]+), ?(.*), ?(DAY|MINUTE|SECOND|HOUR|YEAR)\)/DATETIME_DIFF(\1, \2, '\3')/g"
 export REGEX_DATETIME_TRUNC="s/DATETIME_TRUNC\(([^,]+), ?(DAY|MINUTE|SECOND|HOUR|YEAR)\)/DATE_TRUNC('\2', \1)/g"
 # Add necessary quotes to INTERVAL, e.g. "INTERVAL 5 hour" to "INTERVAL '5' hour"
 export REGEX_INTERVAL="s/interval ([[:digit:]]+) (hour|day|month|year)/INTERVAL '\1' \2/gI"
 # Add numeric cast to ROUND(), e.g. "ROUND(1.234, 2)" to "ROUND( CAST(1.234 as numeric), 2)".
-export PERL_REGEX_ROUND='s/ROUND\(((.|\n)*?)\, /ROUND\( CAST\( \1 as numeric\)\,/g'
+export PERL_REGEX_ROUND='s/ROUND\(((.|\n)*?)(\, [24]\))/ROUND\( CAST\( \1 as numeric\)\3/g'
```
schu, line 12: The regex is rather fragile.

Would it be possible to always `ROUND(CAST(... as numeric), ...)` in the BigQuery SQL code to make the transformation for postgres unnecessary? For example https://github.com/MIT-LCP/mimic-code/blob/main/mimic-iv/concepts/measurement/urine_output_rate.sql#L80 does `CAST`. In the postgres code this leads to `ROUND(CAST(CAST(... as numeric) as numeric), 4)`.

Ideas welcome.

alistairewj: Yeah these regexes are in general terrifying.... since we're starting to come up with non-standard rules for the SQL, we should probably put them somewhere. First thought is the style guide might be a reasonable place (albeit not perfect).

I think the change you're suggesting makes sense, and I'd be happy to get rid of the perl regex, so I'll investigate by bringing this into a branch (tests won't run successfully on external PRs).

alistairewj:
Thanks for the changes (and the other ones as well!). Which OS are you running this on? For the regexes, they are quite brittle as I'd like to support two common flavours of shell:

* BSD sed and bash 3.x (shipped on Mac OS X)
* GNU sed and bash 4.x+ (shipped on Ubuntu)

Trying to allow for this compatibility eliminates the possibility of using some common syntax.

schu:
> Which OS are you running this on?

Fedora 36 with Postgres 14.3 (the latest version in dnf).

schu:
Hey, any news? :)

> I'd be happy to get rid of the perl regex

If you want, I can test and push an alternative patch and drop the regex.

alistairewj:
I got way too distracted by other things and the MySQL container build - but yes especially with people running into issues (e.g. #1393) this should definitely be fixed... making some changes now

alistairewj:
OK this looks to work well - the checks will fail since they're running on a fork which is unprivileged.. it's a TODO of mine to make PRs trigger a privileged workflow rather than trying to run the tests themselves.

I tested it locally and built all the tables successfully (BSD sed over here). @schu can you give convert_bigquery_to_postgres.sh and postgres-make-concepts.sql a run with the latest commits pulled down? If it works for you, then we should be good to merge.

alistairewj:
I'll merge this as I'll test the functionality with gh actions

PR #1366, opened 2022-08-20
Update age.sql

jerryzuo0214:
the variable YEAR was not correctly labeled and should be embeded by single quotes


jerryzuo0214:
Windows10 21H2 operator system

alistairewj:
Same story as the other PR - thanks for the fix, I just confirmed it's correct in the latest code version on main.

PR #1367, opened 2022-08-20
Update icustay_detail.sql

jerryzuo0214:
the functions in this script were misplaced


jerryzuo0214:
Windows10 21H2  operator system

PR #1369, opened 2022-08-20
Update icustay_detail.sql

jerryzuo0214:
parameter YEAR should be embeded by quotes


jerryzuo0214:
Windows10 21H2 operator system

alistairewj:
Thanks for the PR - looks like we did fix this eventually when revamping where the postgres concepts were.

PR #1419, opened 2022-11-08
Add procedures mappings for MIMIC-IV v2.0 - SSSOM format

a-chahin:
This pull request adds two mapping csv files for procedures concepts from the `d_items` and `procedureevents` tables in MIMIC-IV v2.0. The first file `procedureevents_to_snomed.csv`  contains `itemid` to `SNOMED-CT` mappings. The second file `procedureevents_to_omop.csv` contains `itemid` to `OMOP` mappings. 

Both files use the simple standard for sharing ontology mappings `SSSOM` format [@sssom](https://github.com/mapping-commons/sssom)


tompollard:
Looks good, thanks @a-chahin!

- I notice that you have left some of the `author_id` fields empty for the mappings. I assume this is because we inherited those mappings?
- The documentation at https://mapping-commons.github.io/sssom/author_id/ suggests that `author_id` is "Recommended to be a (pipe-separated) list of ORCIDs". I think you are currently using commas to separate lists. Should we swap out the commas for `|`?
- Minor thing, but both files are missing a newline character at the end. This can lead to some tools to ignore the final line. If you are editing the files, we might want to add a newline (just hit return at the end of the final line of each file).

@matentzn hope you don't mind me tagging you here! This file is a mapping for MIMIC-IV concepts. No pressure, but we would appreciate your thoughts on whether anything more should be done to conform to @sssom format.

a-chahin:
> Looks good, thanks @a-chahin!
> 
> * I notice that you have left some of the `author_id` fields empty for the mappings. I assume this is because we inherited those mappings?
> * The documentation at https://mapping-commons.github.io/sssom/author_id/ suggests that `author_id` is "Recommended to be a (pipe-separated) list of ORCIDs". I think you are currently using commas to separate lists. Should we swap out the commas for `|`?
> * Minor thing, but both files are missing a newline character at the end. This can lead to some tools to ignore the final line. If you are editing the files, we might want to add a newline (just hit return at the end of the final line of each file).
> 
> @matentzn hope you don't mind me tagging you here! This file is a mapping for MIMIC-IV concepts. No pressure, but we would appreciate your thoughts on whether anything more should be done to conform to @sssom format.

Thank you @tompollard!

- You are correct. The `author_id` field is empty for concepts that were mapped by the `OMOP` folks.
- Thank you for bringing this to my attention. I'm sorry I missed it.
- I will fix the ORCIDs and add a newline character at the end and update the csv files.

tompollard:
I have (1) added the missing newline and (2) changed the orcid list separator from `,` to `|`.

tompollard (approved):
Thanks @a-chahin !

PR #1455, opened 2022-12-20
Avoid unnecessary dependency on sqlalchemy

bemoody:
`sqlite/import.py` provides a quick and easy way to import MIMIC-IV into SQLite format.  As far as I can see, there's no reason for this script to be dependent on sqlalchemy.

(I'd remove the pandas dependency too, but pandas is actually serving a purpose here by auto-detecting some of the column types.)

Granted, I haven't actually used sqlalchemy and I don't know if it provides any special benefits here.  I also only tested this on the demo (https://physionet.org/content/mimic-iv-demo/1.0/).  Let me know if you want me to do more extensive testing.

It's a minor thing, but it's nice to have one fewer dependency to worry about.

### DIFF
--- a/.github/workflows/sqlite.yml
+++ b/.github/workflows/sqlite.yml
@@ -1,0 +1,39 @@
+name: sqlite demo db build
+on:
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  mimic-iv-sqlite:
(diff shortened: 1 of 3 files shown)


alistairewj (approved):
Looks good - I added a GitHub action to test the build too.

Issue #1408, opened 2022-10-24
Possible inaccuracy in Charlson comorbidity index calculation


tchang1997:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked the online documentation: https://mimic.mit.edu/
    * Checked that your issue isn't already addressed: https://github.com/MIT-LCP/mimic-code/issues?utf8=%E2%9C%93&q=

### Description

I noticed that the SQL scripts for Charlson comorbidity index (CCI) calculation appear to be calculating the age score incorrectly. The [implementation in the repo](https://github.com/MIT-LCP/mimic-code/blob/main/mimic-iv/concepts/comorbidity/charlson.sql) is
```
SELECT
     ...age
  , CASE WHEN age <= 40 THEN 0
    WHEN age <= 50 THEN 1
    WHEN age <= 60 THEN 2
    WHEN age <= 70 THEN 3
    ELSE 4 END AS age_score
```

whereas, based on the [original paper on age-adjusted CCI by Charlson](https://www.sciencedirect.com/science/article/pii/0895435694901295?via%3Dihub), it appears that the correct mapping from age to score would be 
```
SELECT 
     ...age
  , CASE WHEN age <= 50 THEN 0
    WHEN age <= 60 THEN 1
    WHEN age <= 70 THEN 2
    WHEN age <= 80 THEN 3
    ELSE 4 END AS age_score
```
as prescribed in the paper: "Each decade of age over 40 adds 1 point to risk (e.g. 50-59 years, 1 point; 60-69 years, 2 points; 70-79 years, 3 points) and these points for age are added to the score from the Charlson comorbidity index (e.g., 0, 1,2,3, etc.)" This is consistent with other implementations of CCI, e.g. [MDcalc](https://www.mdcalc.com/calc/3917/charlson-comorbidity-index-cci). 

This issue affects the majority of admissions in MIMIC. I reran the updated query on my copy, and this update changed the CCI in 293409 of 454324 (64.6%) of admissions — the CCI in these admissions was previously overestimated by 1. 

If needed, I'm happy to open a PR to patch this.

alistairewj:
Good catch! Would love a PR. Slowly making my way through the backlog, things that make life easy are appreciated :)

alistairewj:
Closed by #1463

PR #1463, opened 2023-01-03
Update charlson.sql for correct age-adjusted CCI score implementation

tchang1997:
Follow-up on #1408 — the Charlson comorbidity score had slightly incorrect age adjustments. I've patched the two `charlson.sql` queries to account for this.


alistairewj:
awesome thanks!

alistairewj (approved):

PR #1467, opened 2023-01-11
feat: Add mimic-iv-note (v2.2) loading of postgres

DopamineLcy:
Add mimic-iv-note (v2.2) loading method of postgres, thank you!


DopamineLcy:
BTW, if I add the constraint of "hadm_id" in "radiology" NOT NULL, an error will occur, which is not consistent with https://mimic.mit.edu/docs/iv/modules/note/radiology/, maybe the data should be checked.
So I remove the NOT NULL constraint of "hadm_id" in "radiology".
Thank you very much.

alistairewj:
This looks great, thank you! Going to merge this in and update any data types that may have changed since we described the data on the website (including the `NOT NULL` constraint like you noted).

PR #1468, opened 2023-01-12
concepts_postgres: updates (on top of `mimiciv_v2.2_updates`)

schu:
A few changes on top of https://github.com/MIT-LCP/mimic-code/tree/mimiciv_v2.2_updates to update the Postgres concepts code. Please see the individual commit messages for details.


alistairewj:
Ah thanks for pulling in those updates, though it's interesting your run sorts `first_day_bg.sql` *after* `first_day_bg_art.sql`, as mine doesn't. `first_day_bg.sql` needs to come before `first_day_bg_art.sql` so I will manually make the fix, but it might be worth taking a look to see why it's happening in the future.

alistairewj:
Nevermind about the above, I forgot I already made them both rely on the `bg` view to avoid that above inconsistency :)

PR #1480, opened 2023-02-08
Add linting of SQL files using sqlfluff

alistairewj:
There is a package for linting SQL files called [sqlfluff](https://docs.sqlfluff.com/en/stable/index.html) ([GitHub link](https://github.com/sqlfluff/sqlfluff)).

I've added a rule file using the simplest core rules plus L019 which enforces leading commas. Importantly, this PR has no functional changes, and only the application of `sqlfluff fix` to the `mimic-iv/concepts` subfolder. Further, some rules were skipped on a file-specific level to ensure that the conversion to PostgreSQL scripts would still work.

I've also added a GitHub action which *should* verify that files are appropriately linted on a PR and annotate errors if they are found. Haven't tested it yet though, so here goes nothing :)

### DIFF
--- a/mimic-iv/concepts/measurement/creatinine_baseline.sql
+++ b/mimic-iv/concepts/measurement/creatinine_baseline.sql
@@ -1,10 +1,13 @@
--- This query extracts the serum creatinine baselines of adult patients on each hospital admission.
+-- This query extracts the serum creatinine baselines of adult patients
+-- on each hospital admission.
 -- The baseline is determined by the following rules:
---     i. if the lowest creatinine value during this admission is normal (<1.1), then use the value
---     ii. if the patient is diagnosed with chronic kidney disease (CKD), then use the lowest creatinine value during the admission, although it may be rather large.
---     iii. Otherwise, we estimate the baseline using the Simplified MDRD Formula:
+--     i. if the lowest creatinine value during this admission is normal (<1.1),
+--          then use the value
+--     ii. if the patient is diagnosed with chronic kidney disease (CKD),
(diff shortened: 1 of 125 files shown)


mimic-iv/concepts/measurement/creatinine_baseline.sql
```diff
@@ -1,69 +1,71 @@
...
---     ii. if the patient is diagnosed with chronic kidney disease (CKD), then use the lowest creatinine value during the admission, although it may be rather large.
---     iii. Otherwise, we estimate the baseline using the Simplified MDRD Formula:
+--     i. if the lowest creatinine value during this admission is normal (<1.1),
+--          then use the value
+--     ii. if the patient is diagnosed with chronic kidney disease (CKD),
+--          then use the lowest creatinine value during the admission,
+--          although it may be rather large.
+--     iii. Otherwise, we estimate the baseline using Simplified MDRD:
 --          eGFR = 186 × Scr^(-1.154) × Age^(-0.203) × 0.742Female
---     Let eGFR = 75. Scr = [ 75 / 186 / Age^(-0.203) / (0.742Female) ] ^ (1/-1.154)
```
tompollard, line 7: This line looks like it has disappeared. Presumably manually edited?

tompollard (commented):
Good to know about this linter!

mimic-iv/concepts/measurement/creatinine_baseline.sql
alistairewj: Yeah it was a redundant comment

PR #1492, opened 2023-02-22
Update load.sql

trostalski:
As this is the load script for the unzipped files, it should be `caregiver.csv` I guess.


tompollard:
Thanks @trostalski, looks good to me!

alistairewj (approved):

PR #1494, opened 2023-02-24
Update postgres-make-concepts.sql

trostalski:
Looks to me like the file `medication/norepinephrine.sql` is missing here. Therefore, the scripts for the dependent tables are failing (and also `score/sofa.sql` which also seems to be dependent). 😄


alistairewj:
Thanks for highlighting this. The postgres-make-concepts script is actually automatically generated by convert_bigquery_to_postgres.sh. The bug is that it skipped adding `medication.norepinephrine` because the check was a regex which incorrectly matched against `medication.norepinephrine_equivalent_dose`. Fixing it in an upcoming PR.

PR #1517, opened 2023-04-01
update load mimic-iv-note into a local postgresql tutorial

yongfanbeta:
https://github.com/MIT-LCP/mimic-code/discussions/1490

### DIFF
--- a/mimic-iv-note/buildmimic/postgres/README.md
+++ b/mimic-iv-note/buildmimic/postgres/README.md
@@ -9,10 +9,10 @@ The scripts in this folder create the schema for MIMIC-IV-Note and load the data
 git clone https://github.com/MIT-LCP/mimic-code.git
 cd mimic-code
 # download data
-wget -r -N -c -np --user <USERNAME> --ask-password https://physionet.org/files/mimiciv/2.2/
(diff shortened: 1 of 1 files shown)


alistairewj:
Hmm is there an advantage in using a second database `mimiciv_note`? For me it seems easier to just use a single database (`mimiciv`) for all the data.

yongfanbeta:
@alistairewj  the main reason is that [mimiciv](https://www.physionet.org/content/mimiciv/2.2/) and [mimic-iv-notes](https://www.physionet.org/content/mimic-iv-note/2.2/) are now stored separately on PhysioNet, if we use a single datebase like mimiciii, you need to update two database on physionet simultaneously, otherwise, when one datebase is change, we need to load all data again including the data don't change.

alistairewj:
Each dataset is stored on a different schema (`mimiciv_hosp`, `mimiciv_note`, `mimiciv_ed`). You definitely don't need to reload data in one schema when creating/loading the other.

yongfanbeta:
oh! I get it. so we only need to change the mimiciv_note data_dir in readme file.

alistairewj:
Yes probably! Feel free to open a new PR if there are modifications you think would be useful given our discussion.

PR #1529, opened 2023-04-28
DuckDB MIMIC-III concepts implementation

### LINKED ISSUE #1495: MIMIC-III Concepts for Sqlite3 or DuckDB (probably now DuckDB)
### Prerequisites

### Description

Filing this because I am considering adding it: Create a set of scripts that will add the MIMIC-III concepts views to the Sqlite3 version of the database.

Has anyone tried this before, and if so any encountered difficulties? I'm having some initial success just with regexp-ing the Postgres scripts, though that may or may not be a practical solution for the whole task. Sqlite does not have functions/procedures, so some things might have to be done with python functions--which would preclude embedding them into materialized views (pretty sure), however, simply building the concepts entities as static tables may be an acceptable solution for the kinds of use cases you'd want to use this data and Sqlite for. Or maybe regexping the Postgres/BigQuery SQL will get me to views in Sqlite3 dialect.

SphtKr:
These changes expand on #1489 / #1239 by adding a Python version that ingests the data _and_ creates the concepts views in the DuckDB database. The python version requires only the duckdb python package and not the command-line executable. Also requires [sqlglot](https://github.com/tobymao/sqlglot).

Run `python import-duckdb.py --help` for usage, it's pretty simple.

Tested on the MIMIC-III demo and the full version (1.4)... it seems to work, though I'm not sure if there's a better way to test (i.e. compare actual values... I have the Postgres version running in a container if so).

Resolves #1495


mimic-iii/buildmimic/duckdb/import_duckdb.py
```diff
@@ -0,0 +1,378 @@
...
+    # work... So, here, read the tables already in the destination schema from
+    # the DB and add those tablenames to tables_in_schema?
+
+    if make_concepts:
+        connection = duckdb.connect(output_db)
+        print("Connected to duckdb...")
+
+        print("Creating tables...")
+
+        # ccs_dx is an outlier...this is adapted from the BigQuery version...
```
alistairewj, line 306: why is it an outlier?

mimic-iii/buildmimic/duckdb/import_duckdb.py
```diff
@@ -0,0 +1,378 @@
...
+
+            for key in concept_name_map:
+                if schema_name is not None:
+                    if "schema" not in concept_name_map[key]:
+                        tables_in_schema.add(key.lower())
+
+                #cProfile.run('...')
+                #print(f"Making view {key}...")
+                db = concept_name_map[key].get("db", "bigquery")
+                if db == "duckdb":
```
alistairewj, line 358: I think this is unnecessarily abstract. You might as well just hard-code the make_duckdb_query function to do something special for icustay_hours. Or, alternatively, have the subfunction contain a hard-coded list of queries which it parses as duckdb, that way you can expand it to more queries in the future if you plan to move more over

mimic-iii/buildmimic/duckdb/import_duckdb.py
```diff
@@ -0,0 +1,378 @@
...
+
+def _make_duckdb_query_duckdb(qname: str, qfile: str, conn, schema: str = None):
+    with open(qfile, "r") as fp:
+        sql = fp.read()
+
+        if schema is not None:
+            sql = _duckdb_rewrite_schema(sql, schema)
+
+        try:         
+            conn.execute(f"CREATE VIEW {qname} AS " + sql)
```
alistairewj, line 213: missing a drop view if exists -> will be fixed by harmonizing duckdb/bigquery subfunctions

mimic-iii/buildmimic/duckdb/import_duckdb.py
```diff
@@ -0,0 +1,378 @@
...
+def _make_duckdb_query_bigquery(qname: str, qfile: str, conn, schema: str = None):
+    _multischema_trunc_re = re.compile("\"physionet-data\.mimiciii_\w+\.")
+    
+    #TODO: better answer here? should only hit ccs_dx.sql!
+    _too_many_backslashes_re = re.compile("\\\\([\[\.\]])") 
+
+    with open(qfile, "r") as fp:
+        sql = fp.read()
+        sql = re.sub(_too_many_backslashes_re, '\\$1', sql) 
+        try:
```
alistairewj, line 182: i would rewrite this merging the bigquery/duckdb subfunctions into one, and having an if statement which calls the sql modification if needed, and then continuing on as usual. this would avoid mild divergences in behaviour observed e.g. forgetting to drop the view for duckdb but doing it for bigquery

mimic-iii/buildmimic/duckdb/import_duckdb.py
```diff
@@ -0,0 +1,378 @@
...
+            raise e
+        for st in sql_list:
+            sql = re.sub(_multischema_trunc_re, "\"", st)
+
+            if schema is not None:
+                sql = _duckdb_rewrite_schema(sql, schema)
+
+            conn.execute(f"DROP VIEW IF EXISTS {qname}")
+            try:         
+                conn.execute(f"CREATE VIEW {qname} AS " + sql)
```
alistairewj, line 195: What about creating either materialized views or simply tables? A view would rerun the code over and over (and be slow). I think most users will be happy with trading off the disk space for faster queries, so I'd say the default should be to create tables and if a `--view` argument is passed it creates views instead (or two arguments, `--make-concepts` and `--make-concepts-as-views`).

alistairewj (changes requested):
Some initial thoughts from a scan of the code.

I also think there are a lot of if/else statements littered throughout, but haven't thought about it enough to simplify that yet.

mimic-iii/buildmimic/duckdb/README.md
```diff
@@ -102,6 +105,66 @@ The script will print out progress as it goes.
 Be patient, this can take minutes to hours to load
 depending on your computer's configuration.
 
+## Python script method (`import_duckdb.py`)
+
+This method does not require the DuckDB executable, the DuckDB Python
```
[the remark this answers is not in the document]
SphtKr: Correct--the shell script requires the executable to be in the path. The DuckDB Python module (loaded through pip/PyPi) just loads Python modules (doesn't even add a version in a venv's `bin` dir, for instance).

mimic-iii/buildmimic/duckdb/duckdb_add_tables.sql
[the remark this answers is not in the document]
SphtKr: I can try putting this through sqlglot--I think I made that SQL script before I started using that. At present, there is a bug (referenced in the comment in that file) that causes creation of the table to fail with the primary key in place (index creation causes OOM)... so until that's fixed the table creation will need at least that tweak.

mimic-iii/buildmimic/duckdb/concepts/icustay_hours.sql
```diff
@@ -0,0 +1,21 @@
...
+SELECT
+  ah.icustay_id,
+  unnest(ah.hr) as hr, 
+  /* endtime now indexes the end time of every hour for each patient */
+  unnest(list_transform(hr, ahr -> ah.endtime + ahr*INTERVAL 1 hour)) AS endtime
+  --ah.endtime+ hr*INTERVAL 1 hour as endtime
+FROM all_hours AS ah
+ORDER BY
+  ah.icustay_id NULLS LAST
+limit 20
```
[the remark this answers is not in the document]
SphtKr: 😱 probably.

mimic-iii/buildmimic/duckdb/import_duckdb.py
SphtKr: Probably valid.... these two were originally far more divergent but are indeed quite similar now.

mimic-iii/buildmimic/duckdb/import_duckdb.py
SphtKr: Probably a good idea. Looks like DuckDB doesn't have materialized views (yet), but for this use case tables are perfectly valid.

mimic-iii/buildmimic/duckdb/import_duckdb.py
SphtKr: Yes, comment should probably be more specific. For the record, 1. because it's the only non-query-based table (that gets loaded with data) in the concepts build process, 2. the Postgres DB puts it in the public schema instead of `mimiciii`... which I am guessing is because it's also applicable to the MIMIC-IV concepts?

mimic-iii/buildmimic/duckdb/import_duckdb.py
SphtKr: Yes.... earlier I was pulling some queries from the BigQuery version and others from the Postgres version, and Postgres was actually the default...later I got them all to work translated from BigQuery except this one...which now looks odd.

mimic-iii/buildmimic/duckdb/import_duckdb.py
```diff
@@ -0,0 +1,378 @@
...
+                    sql = fp.read()
+                    connection.execute(sql)
+
+            print("Running checks...")
+    
+            with open(os.path.join(mimic_code_root, 'mimic-iii','buildmimic','duckdb','duckdb_checks.sql'), 'r') as fp:
+                sql = fp.read()
+                result = connection.execute(sql).fetchall()
+                for row in result:
+                    print(f"{row[0]}: {row[2]} records ({row[1]} expected) - {row[3]}")
```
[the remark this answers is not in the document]
SphtKr: I was thinking about `--skip-checks`... should I just make `--run-checks` (off by default)? Don't love the idea of another set of static values...though I suppose they wouldn't change much.

mimic-iii/buildmimic/duckdb/duckdb_add_tables.sql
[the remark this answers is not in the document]
alistairewj: It would be nice to have a json used to generate the DDL for postgres/bigquery/mysql/duckdb - but it also seems like a separate PR to duckdb concepts

I suppose the index OOM bug is a duckdb bug?

mimic-iii/buildmimic/duckdb/import_duckdb.py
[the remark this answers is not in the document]
alistairewj: I ended up with a `validate.sql` and a `validate_demo.sql` with MIMIC-IV fwiw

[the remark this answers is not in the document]
alistairewj: Though again, the row counts are something that screams "this should be a json somewhere else in the repo", rather than hard-coded into all these destined to rot SQL files

alistairewj:
Seems like there are a few valid changes, feel free to update in your own time and just ping me with a re-review when you think it's ready

mimic-iii/buildmimic/duckdb/duckdb_add_tables.sql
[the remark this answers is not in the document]
SphtKr: > I suppose the index OOM bug is a duckdb bug?

Yep, it's [resolved in their repo](https://github.com/duckdb/duckdb/issues/6668) but not released yet.

SphtKr:
I have a very interesting problem.

I noticed that the DuckDB version here is producing _integer_ results for `DATETIME_DIFF`-type operations, e.g. `icustay_detail.los_icu`, but the PostgreSQL version (implemented in `postgres-functions.sql`) returns _fractional_ values, e.g. 3.14 days vs. 3 days (the DuckDB version is in fact a `floor` of the Postgres version, it appears). So I was going to fix that...

Then, since I am actually converting the queries from the BigQuery versions, I consulted the [BigQuery documentation for `DATETIME_DIFF`](https://cloud.google.com/bigquery/docs/reference/standard-sql/datetime_functions#datetime_diff) which says "Returns the whole number of specified part intervals..." so it looks like the BigQuery version of the concepts may differ from the Postgres versions in this area--though I don't have easy access to the BigQuery version to test.

It looks like I can do either in the DuckDB version by changing how the SQL is translated... I may even  put in an option to do either... but which one _should_ it be, if there is a "should"? If nothing else I wanted to bring this up if it's an issue to have the discrepancy between the two current versions.

alistairewj:
Definitely an inconsistency, good spot. I can confirm BigQuery is giving integers after calculating a floor. Immediately I think the easiest fix is to add a `FLOOR()` call to each of the postgresql functions which emulate BigQuery behaviour (which can be done in a separate PR). I'll raise an issue for that. I think there are a few queries where this would obviously impact the output as it's generated from a `DATETIME_DIFF` call (age, icustay_detail), and some queries where it's more subtle (e.g. urine_output_rate).

For duckdb I think return whole integers is fine.

alistairewj:
closing this as almost all the commits were merged in #1689 a while ago - thanks again!

Issue #684, opened 2020-02-04
Docker image error:  LOG: invalid length of startup packet


khalid-hussain:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked the online documentation: https://mimic.physionet.org/about/mimic/
    * Checked that your issue isn't already addressed: https://github.com/MIT-LCP/mimic-code/issues?utf8=%E2%9C%93&q=

### Description

Greetings,

I'm having a little trouble setting up the docker image. I've followed the guide here: https://github.com/MIT-LCP/mimic-code/tree/master/buildmimic/docker

My System:

```
------------------
System Information
------------------
Operating System: Windows 10 Pro 64-bit (10.0, Build 18363) (18362.19h1_release.190318-1202)
    Language: English (Regional Setting: English)
Processor: Intel(R) Core(TM) i7-3520M CPU @ 2.90GHz (4 CPUs), ~2.9GHz
Memory: 16384MB RAM
Available OS Memory: 16080MB RAM
Page File: 14726MB used, 10058MB available
Windows Dir: C:\WINDOWS
```

My current folder tree at (D:\):

```
.
|-- mimic_code
|   |-- benchmark
|   |-- buildmimic
|   |-- concepts
|   |-- notebooks
|   |-- tests
|   `-- tutorials
|-- mimic_data
|   `-- csv
`-- mimic_pgdata
```

The folder at `\mimic-data\csv` contains the MIMIC-III `.csv.gz` data files. The folder `mimic_pgdata` is empty at the time of running the container. It's empty even after running the container. I assume this is because I do not execute any commands after the container is created.

The command I used (using the Git Bash terminal):

```
docker run \
--name mimic \
-p 5433:5432 \
-e BUILD_MIMIC=1 \
-e POSTGRES_PASSWORD=postgres \
-e MIMIC_PASSWORD=mimic \
-v //d/mimic_data/csv:/mimic_data \
-v //d/mimic_pgdata:/var/lib/postgresql/data \
-d postgres/mimic
```

My result from the Docker Dashboard:

```
PostgreSQL Database directory appears to contain a database; Skipping initialization
2020-02-04 15:23:25.966 UTC [1] LOG:  starting PostgreSQL 12.1 (Debian 12.1-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
[...]

PR #1635, opened 2023-09-06
remove stale/broken docker scripts, see #684

alistairewj:
Long discussion in #684

Summary is that the scripts are broken, and rather than have broken scripts, we should remove the build


tompollard (approved):
Looks good to me, thanks!

PR #1639, opened 2023-09-12
fixing wrong mimic-iv folder name

fvillena:


tompollard:
good catch, thanks @fvillena!

PR #1659, opened 2023-10-18
Fix constraint.sql for mimic-iv-ed PostgreSQL

SimonUzL:
The script constraint.sql for building MIMIC IV ED with a PostgreSQL database seems to contain the wrong schema name.


alistairewj (approved):
whoops! thanks for the fix.

PR #1417, opened 2022-11-08
Add vital signs mappings for MIMIC-IV v2.0 - SSSOM format

a-chahin:
This pull request adds two mapping csv files for vital signs concepts from the `d_items` and `chartevents` tables in MIMIC-IV v2.0. The first file `chartevents_to_loinc.csv`  contains `itemid` to `LOINC` mappings. The second file `chartevents_to_omop.csv` contains `itemid` to `OMOP` mappings. 

Both files use the simple standard for sharing ontology mappings `SSSOM` format [@sssom](https://github.com/mapping-commons/sssom)

PR #1660, opened 2023-10-18
Adds initial mimic-northwestern derived concept mappings

danamouk:
This pull request adds the mimic-northwestern derived concepts mappings to the mimic code repository to link as part of the mimic-website.


tompollard:
Thanks Dana. It looks like there is a significant overlap with our earlier work to create these vital sign mappings for MIMIC in SSSOM format (https://github.com/MIT-LCP/mimic-code/pull/1417). Please could you review that pull request, and then make your amendments and additions on top (instead of introducing a competing set of mappings)?

danamouk:
Thanks @tompollard, I have updated the changes to the vital signs in #1417 to include the mimic and northwestern mappings to loinc and omop. I think we can close this pull request if everything looks captured as part of #1417!

PR #1674, opened 2023-11-24
DATETIME_SUB is not supported by PG

tiansiyuan:


alistairewj:
This is intentional as the postgres functions script creates this needed function: https://github.com/MIT-LCP/mimic-code/blob/16c8cda4f06da57cabad6b6b971a298f3de04f62/mimic-iv/concepts_postgres/postgres-functions.sql

I have overhauled the concept generation for postgres though, which I'll PR soon

PR #1680, opened 2023-12-11
nsaid & acei medication

Juan-L-Dominguez:
SQL code to get nsaid and acei prescriptions


alistairewj (approved):
seems reasonable to me, thanks!

PR #1687, opened 2023-12-31
add validate.sql for mimic-iv-note

Jin-Zhang-Yaoguang:
add mimic-iv-note validate.sql


alistairewj (approved):
I am not sure why discharge detail has so few rows compared to discharge but this is consistent with the data on PhysioNet so looks good

PR #1689, opened 2024-01-06
Add duckdb build/concepts and use SQLGlot to convert BigQuery SQL into other dialects

alistairewj:
A few things in this PR, mainly to get the concepts to work with duckdb. This builds off #1529 / @SphtKr. However #1529 has the conversion for MIMIC-III, whereas here I've converted MIMIC-IV. It should be straightforward to adapt for MIMIC-III, but I can't explain the parsing errors I'm getting when running it over that folder, so I've left it for a future PR.

- Added build scripts for mimic-iii in duckdb
- Added build scripts for mimic-iv in duckdb
- Added a mimic_utils package that converts files/folders between SQL dialects.
- Use this package to re-convert the postgresql scripts, and remove the old bash script system for MIMIC-IV
- Moved the mapping folder from mimic-iv/concepts/mapping to mimic-iv/mapping

I also overhauled the READMEs a bit for clarity.


mimic-iv/buildmimic/duckdb/README.md
```diff
@@ -43,10 +43,11 @@ e.g. `/usr/local/bin`.
 
 ### Download MIMIC-IV files
 
-Download the CSV files for [MIMIC-IV v2.0](https://physionet.org/content/mimiciv/2.0/)
+Download the CSV files for [MIMIC-IV](https://physionet.org/content/mimiciv/)
 by any method you wish.
+These instructionds were tested with MIMIC-IV v2.2.
```
tompollard, line 48: "instructionds" -> "instructions"

mimic-iv/concepts_duckdb/README.md
```diff
@@ -0,0 +1,16 @@
+# duckdb concepts
+
+This folder has SQL compatible with [DuckDB](https://duckdb.org/).
+These concepts were generated automatically from the BigQuery SQL dialect using the [sqlglot](https://sqlglot.com/) package.
+If you would like to contribute a correction, it should be for the corresponding file in the concepts folder.
```
tompollard, line 5: I think the following sentence is saying that corrections should be made to the BigQuery SQL dialect? 

`If you would like to contribute a correction, it should be for the corresponding file in the concepts folder.`

If so (which I think is best), it might be worth making the statement more explicit (e.g. "in the BigQuery concepts folder").

tompollard (approved):
Thanks, looks good to me, with the caveat that I haven't run the code! (and that breaking this into smaller PRs would be nicer). I added a couple of notes about typos but all very minor, so feel free to go ahead and merge. The `sqlglot` package looks very cool. 

How much manual tweaking of the outputs is needed? 

Ideally I think we would maintain a single source format (psql/bigquery) and then create other dialects in a GitHub workflow. My preference would be to have dialects on separate branches, rather than trying to maintain them all on main.

PR #1716, opened 2024-02-26
feat: add avg statistics for first day parameters

aegis301:
The first day derived tables are probably some of the communities favourites. I added some more statistics to them for easier usage. Feel free to suggest additional statstics of interest.


alistairewj:
these changes need to be done on concepts/ - as the postgres concepts are automatically generated

(apologies for not reviewing in a timely manner!)

PR #1718, opened 2024-03-04
change mimic data into sqlite

Linyi-Chen:
This contribution focuses on the process of importing the extensive MIMIC (Medical Information Mart for Intensive Care) dataset into a structured SQLite database. The primary goal is to transform the raw MIMIC data, which is distributed across multiple CSV files, into a more accessible and queryable format within a SQLite database. This transformation facilitates easier data exploration, analysis, and research for healthcare professionals and data scientists alike.


alistairewj:
The SQLite python/shell scripts work (now) - so we will stick with those for now, but thank you for the contribution!

Issue #1736, opened 2024-04-18
icustay_hourly.sql run error


liuyao123456:
### Description
Error: function unnested (integer) does not exist
LINE 17: CROSS JOIN UNNEST(all_hours.hrs) AS _t(hr_unnested)
                    ^
HINT: There is no function that matches the specified name and parameter type. You may need to add explicit type conversion.

HuilingHU:
you can try:
the last line of the icustay_hourly.sql----> CROSS JOIN UNNEST(array[all_hours.hrs]) AS _t(hr_unnested)

theompm:
Hi i got the whole pipeline run by editing:
mimic-iv/concepts_postgres/demographics/icustay_hourly.sql

```sql
-- THIS SCRIPT IS AUTOMATICALLY GENERATED. DO NOT EDIT IT DIRECTLY.
DROP TABLE IF EXISTS mimiciv_derived.icustay_hourly; CREATE TABLE mimiciv_derived.icustay_hourly AS
/* This query generates a row for every hour the patient is in the ICU. */ /* The hours are based on clock-hours (i.e. 02:00, 03:00). */ /* The hour clock starts 24 hours before the first heart rate measurement. */ /* Note that the time of the first heart rate measurement is ceilinged to */ /* the hour. */ /* this query extracts the cohort and every possible hour they were in the ICU */ /* this table can be to other tables on stay_id and (ENDTIME - 1 hour,ENDTIME] */ /* get first/last measurement time */
WITH all_hours AS (
  SELECT
    it.stay_id, /* round the intime up to the nearest hour */
    CASE
      WHEN DATE_TRUNC('HOUR', it.intime_hr) = it.intime_hr
      THEN it.intime_hr
      ELSE DATE_TRUNC('HOUR', it.intime_hr) + INTERVAL '1 HOUR'
    END AS endtime, /* create integers for each charttime in hours from admission */ /* so 0 is admission time, 1 is one hour after admission, etc, */ /* up to ICU disch */ /*  we allow 24 hours before ICU admission (to grab labs before admit) */
    GENERATE_SERIES(-24, CAST(CEIL(EXTRACT(EPOCH FROM it.outtime_hr - it.intime_hr) / 3600.0) AS INT)) AS hrs /* noqa: L016 */
  FROM mimiciv_derived.icustay_times AS it
)
SELECT
  stay_id,
  CAST(all_hours.hrs AS BIGINT) AS hr,
[...]

Issue #1734, opened 2024-04-14
some questions about mimic_derived when i tried to install icustay hourly,sofa and sepsis3.


njmuyxd:
When i tried to install mimic_derived tables,most of them can be installed correctly,but some of them didn't appear in my navicat like icustay hourly ,sofa, and sepsis3.
icustay_hourly.sql error is the grammar error besides 'CAST',how can i fix it?

alistairewj:
Can you provide more detail on the error and what database system you are using? Which specific tables fail? If it's a Postgres database make sure you use the postgres version of the concepts.

liuyao123456:
> Can you provide more detail on the error and what database system you are using? Which specific tables fail? If it's a Postgres database make sure you use the postgres version of the concepts.

Error: function unnested (integer) does not exist
LINE 17: CROSS JOIN UNNEST(all_hours.hrs) AS _t(hr_unnested)
^
HINT: There is no function that matches the specified name and parameter type. You may need to add explicit type conversion.

liuyao123456:
> When i tried to install mimic_derived tables,most of them can be installed correctly,but some of them didn't appear in my navicat like icustay hourly ,sofa, and sepsis3. icustay_hourly.sql error is the grammar error besides 'CAST',how can i fix it?

please run this icustay_hourly.sql:
with all_hours as
(
select
  it.stay_id

  -- ceiling the intime to the nearest hour by adding 59 minutes then truncating
  -- note thart we truncate by parsing as string, rather than using DATETIME_TRUNC
  -- this is done to enable compatibility with psql
  , PARSE_DATETIME(
      '%Y-%m-%d %H:00:00',
      FORMAT_DATETIME(
        '%Y-%m-%d %H:00:00',
          DATETIME_ADD(it.intime_hr, INTERVAL '59' MINUTE)
  )) AS endtime

  -- create integers for each charttime in hours from admission
  -- so 0 is admission time, 1 is one hour after admission, etc, up to ICU disch
  --  we allow 24 hours before ICU admission (to grab labs before admit)
  , ARRAY(SELECT * FROM generate_series(-24, CEIL(DATETIME_DIFF(it.outtime_hr, it.intime_hr, 'HOUR')))) as hrs

  from mimiciv_derived.icustay_times it
)
SELECT stay_id
, CAST(hr AS bigint) as hr
, DATETIME_ADD(endtime, interval '1' hour * CAST(hr AS bigint)) as endtime
FROM all_hours
[...]

PR #1744, opened 2024-05-01
fix icustay_hourly: function unnest(integer) does not exist

BuiHoangTu:
all_hours.hrs is a row of integer already, so we don't need nor able to use unnest in this case.


BuiHoangTu:
fix #1736 and partially #1734

ZhouQiuyu (approved):
It did solve "function unnest(integer) does not exist" when generating icustay_hourly.

PR #1745, opened 2024-05-02
Adding data preprocessing code for ITQM 2023 paper "Generation of Radiology Findings in Chest X-Ray by Leveraging Collaborative Knowledge"

DanuManuela:
This pull request is to contribute code associated with the **ITQM 2023** paper "**Generation of Radiology Findings in Chest X-Ray by Leveraging Collaborative Knowledge**". The experiments conducted in this paper were based on the publicly available MIMIC-CXR dataset. 

The code we want to publish is the one utilized for preparing the data for the experiments.

The paper is available at this [link](https://www.sciencedirect.com/science/article/pii/S1877050923008529).

### DIFF
--- a/mimic-iv-cxr/findings_generation/README.md
+++ b/mimic-iv-cxr/findings_generation/README.md
@@ -1,0 +1,42 @@
+# Generation of Radiology Findings in Chest X-Ray by Leveraging Collaborative Knowledge
+
+<p align="justify">
+This repository contains the code utilized for preparing the data used in the paper "Generation of Radiology Findings in Chest X-Ray by Leveraging Collaborative Knowledge". The paper was presented and published during the Tenth International Conference on Information Technology and Quantitative Management (ITQM 2023). The concepts and information presented in this paper are based on research results that are not commercially available. Future commercial availability cannot be guaranteed. 
+</p>
+
+The paper is available at this [link](https://www.sciencedirect.com/science/article/pii/S1877050923008529).
+
+<p align="justify">The proposed methodology for generating radiology reports follows a two-step approach:</p>
+
+
+1. <p align="justify">Abnormalities detection and classification in chest X-rays (the code is not available since the model was trained on a private dataset).</p>
+2. <p align="justify">Radiology Findings generation through LLM prompt-tuning (the data pre-processing code is available in this repository since the LLM was trained on the publicly availabele MIMIC-CXR dataset).</p>
+
+
+
+## Data pre-processing
+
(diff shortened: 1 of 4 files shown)


tompollard:
Thanks for the contribution! This pull request makes me realise that we don't have a good way of managing the code that underpins published papers.

I'm wondering whether we should either:

1) Have a papers subfolder in this repository, along with an index of papers (with link, citation, etc) in a markdown document.
2) Instead, ask for the code for papers to be added to a separate repository, and just include an index of papers in this repository.

I would be interested in hearing your thoughts on this. @DanuManuela @alistairewj @bemoody

DanuManuela:
> Thanks for the contribution! This pull request makes me realise that we don't have a good way of managing the code that underpins published papers.
> 
> I'm wondering whether we should either:
> 
> 1. Have a papers subfolder in this repository, along with an index of papers (with link, citation, etc) in a markdown document.
> 2. Instead, ask for the code for papers to be added to a separate repository, and just include an index of papers in this repository.
> 
> I would be interested in hearing your thoughts on this. @DanuManuela @alistairewj @bemoody

DanuManuela:
I am more inclined towards the first option. It just seems easier to have a dedicated folder for papers within the repository, thus keeping all the relevant materials in one place. However, I'm totally open to option two as well.

tompollard:
Thanks @DanuManuela, I basically agree that it would be nice to have everything in one place, but when I think about this a little more my preference is option 2 because:

- Even if we want to have everything in one place, most people will probably end up sharing elsewhere, so this goal is difficult to achieve.
- If we do have everything in one place (i.e. this repo), we are going to struggle with maintenance. My guess is that people will start opening issues and bug fixes to the code. We probably aren't the right people to address those.

Hopefully Benjamin and Alistair can add their thoughts (plus anyone who is interested!).

alistairewj:
Yeah this is really great and thanks for contributing! With that said, I will add that I also think you should make your own repo for this. We can still include a reference here to give your work visibility. Having your own repo will help keep issues related to your work in the separate repo (so you don't get overloaded by mimic code issues, unless you choose to be!), allow citations to properly credit you, and in general it removes the need for us to standardize contributions, which would be very hard indeed. These thoughts are in addition to the ones mentioned above.

One thing that could be nice is a semi-standardized table for inclusion in your README that makes it easier to understand the code. Just simple things like SQL dialect, etc, could be useful. 

So the suggested plan would be;


(1) create your own repository with the code
(2) (optional) mint a doi with Zenodo
(3) contribute a link to that repo / DOI in an index file here (n.b. I had dreams of doing this with MIMIC-CXR and an awesome MIMIC-CXR markdown file but to date these dreams have not materialized).

PR #1779, opened 2024-08-13
Fix bug: Columns dislocation in the 'emar' table.

Scott1743:
Columns dislocation due to duplicate column names when importing the 'emar' table with Mysql. so update the load.sql.

### DIFF
--- a/mimic-iv/buildmimic/mysql/load.sql
+++ b/mimic-iv/buildmimic/mysql/load.sql
@@ -339,7 +339,7 @@ LOAD DATA LOCAL INFILE 'emar.csv' INTO TABLE emar
    FIELDS TERMINATED BY ',' ESCAPED BY '' OPTIONALLY ENCLOSED BY '"'
    LINES TERMINATED BY '\n'
    IGNORE 1 LINES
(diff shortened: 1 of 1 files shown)


Scott1743:
When importing data with MySql, I got this following error data:
image

alistairewj:
Hmm is the provider ID in the raw data you tried to load?

Scott1743:
Emm... I mean these columns of data are out of place.
image
When I looked at the source code, I realized that this line of code might be written incorrectly
image

alistairewj:
Yes I agree! What's missing is the `enter_provider_id`, the column just before the first shift. This column was added in v2.2, and should look like a string of the format `P[A-Z0-9]+` (a P followed by 4-6 alphanumeric characters). Is it in the raw CSV you tried to load? Are you trying to load an earlier version with the existing scripts, or do the existing scripts need to be updated for v2.2? Not sure when we last updated the mysql scripts.

Scott1743:
The existing MySQL script needs to be updated, but after I deleted the duplicate field and imported it, the Emar table looks good.
image

alistairewj (approved):

alistairewj:
Ah got it thanks. I was a bit confused because your initial screenshot doesn't have the provider ID column, so I wasn't sure if you were using the latest version of the data. But obviously there should only be one entry for `enter_provider_id`! Interesting that the MySQL CI build hasn't been failing...

PR #1831, opened 2024-11-22
Utilization pre/post SQL queries

KevinKloiber:
Prior and post-discharge utilization might be helpful or interesting for researchers and their studies. 

While the inpatient count is pretty straightforward, it should be noted that ED visits can only be identified using the transfer table, since the ED table was only included in later years and does thus not account for ED visits that occurred in early years (anchor_year_group 2008-2010). So simply using the ED table that is provided by MIMIC might lead to a miscalculation of utilization, especially in prior years. I also added two post-discharge variables that separate ED visits by ED visits with admission and without admission.

### DIFF
--- a/mimic-iv/concepts/utilization/post_utilization.sql
+++ b/mimic-iv/concepts/utilization/post_utilization.sql
@@ -1,0 +1,155 @@
+-- -------------------------------------------------------------------------
+-- Query: Post-Discharge Healthcare Utilization with ED Visit Classification
+-- Description:
+--   This query calculates the number of Emergency Department (ED) visits and
+--   inpatient readmissions within a specified time window after each hospital
+--   discharge. Additionally, it classifies ED visits based on whether they
+--   resulted in an admission ("with admission") or not ("without admission").
+--   This helps assess the patient's healthcare utilization following
+--   discharge from the hospital, with a distinction between different types of ED visits.
+--
+-- Instructions:
+--   - Set the desired time window in days by modifying the `time_window_days`
+--     value in the `parameters` CTE (Common Table Expression).
+--   - The default time window is 30 days.
+--
+-- Notes:
+--   - The query uses the `admissions` table for inpatient readmissions and the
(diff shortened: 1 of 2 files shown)


Chessing234:
utilization pre/post queries — a one-line description of the cohort definition in the header would help reuse.

Chessing234 (changes requested):
i had asked about the cohort definition earlier; having run these through the transpiler there's a blocking issue as well.

both queries use `TIMESTAMP_ADD`/`TIMESTAMP_SUB` with the interval amount coming from a column (`INTERVAL p.time_window_days DAY`). the repo's transpiler can't convert that form — running it over `post_utilization.sql` for postgres gives:

    AND na.admittime <= TIMESTAMP_ADD(ca.dischtime, p.time_window_days, DAY)

which isn't a postgres function. grepping mimic-iv/concepts_postgres and mimic-iii/concepts_postgres on main returns no hits for `TIMESTAMP_ADD` at all, so there's no udf backing it either — the generated file wouldn't run.

`DATETIME_ADD` does convert cleanly, including with a column as the amount:

    DATETIME_ADD(dischtime, INTERVAL p.w DAY)  ->  dischtime + p.w * INTERVAL '1' DAY

and that's what the other 20 concepts in mimic-iv/concepts use. switching the four call sites over should be enough.

separately, the pr only adds the bigquery sources. new concepts also need their transpiled postgres and duckdb copies committed and an entry in postgres-make-concepts.sql and concepts_duckdb/duckdb.sql, otherwise `generated-up-to-date` flags the missing files and nothing builds the tables.

one design question rather than a defect: the `parameters` cte with a hardcoded 30-day window and an instruction to edit the value makes these parameterised queries rather than fixed derived tables, which is a bit different from everything else under concepts/. i don't have a strong view on whether that's fine or whether they'd sit better in the cookbook — worth a maintainer's call before you spend time on the dialect copies.

KevinKloiber:
Sorry, mislicked on re-request review.

PR #1847, opened 2025-01-15
Fix: use CONCAT with CAST to create interval for DATETIME_ADD

flybitcone:
Change the interval and CROSS JOIN function

### DIFF
--- a/mimic-iv/concepts_postgres/demographics/icustay_hourly.sql
+++ b/mimic-iv/concepts_postgres/demographics/icustay_hourly.sql
@@ -1,22 +1,38 @@
 -- THIS SCRIPT IS AUTOMATICALLY GENERATED. DO NOT EDIT IT DIRECTLY.
-DROP TABLE IF EXISTS mimiciv_derived.icustay_hourly; CREATE TABLE mimiciv_derived.icustay_hourly AS
-/* This query generates a row for every hour the patient is in the ICU. */ /* The hours are based on clock-hours (i.e. 02:00, 03:00). */ /* The hour clock starts 24 hours before the first heart rate measurement. */ /* Note that the time of the first heart rate measurement is ceilinged to */ /* the hour. */ /* this query extracts the cohort and every possible hour they were in the ICU */ /* this table can be to other tables on stay_id and (ENDTIME - 1 hour,ENDTIME] */ /* get first/last measurement time */
+DROP TABLE IF EXISTS icustay_hourly; CREATE TABLE icustay_hourly AS
+-- This query generates a row for every hour the patient is in the ICU.
+-- The hours are based on clock-hours (i.e. 02:00, 03:00).
+-- The hour clock starts 24 hours before the first heart rate measurement.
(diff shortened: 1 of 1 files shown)


flybitcone:
The check is not pass may be due to the Python version in the test environment.

Pls check the error in detain:
Run actions/setup-python@v2
Version 3.7 was not found in the local cache
Error: Version 3.7 with arch x[6](https://github.com/MIT-LCP/mimic-code/actions/runs/12787737612/job/35661515682?pr=1847#step:3:7)4 not found

Thanks.

Chessing234 (commented):
this file is marked auto-generated — usually you want to fix the bigquery source and re-transpile rather than hand-edit concepts_postgres. also the patch drops the `mimiciv_derived.` schema and brings in `GENERATE_ARRAY` / `DATETIME_ADD`, which won't run as plain postgres. if the goal is interval construction, maybe keep `GENERATE_SERIES` + `+ INTERVAL '1' HOUR` like the current transpiled version?

Chessing234:
concat+cast for intervals is the portable fix. does this also clear the same pattern elsewhere in the file?

Chessing234 (changes requested):
i had asked above whether this clears the same pattern elsewhere — having actually run it, i think the branch needs redoing rather than extending, for a few concrete reasons.

the change only touches `mimic-iv/concepts_postgres/demographics/icustay_hourly.sql`, which carries the "AUTOMATICALLY GENERATED. DO NOT EDIT IT DIRECTLY" header. it's produced from `mimic-iv/concepts/demographics/icustay_hourly.sql` by the transpiler, so the next regeneration drops this edit entirely. i checked the whole folder: 65 of the 67 sql files there are generated, the only hand-written ones being postgres-make-concepts.sql and postgres-concept-index.sql. concepts_postgres/README.md also says to use the transpiler rather than hand-translating.

the edit also swaps BigQuery-only functions into the postgres output — `GENERATE_ARRAY`, `DATETIME_DIFF`, `DATETIME_ADD`. grepping mimic-iv/concepts_postgres and mimic-iii/concepts_postgres on main returns no hits for any of those; the transpiler converts them (GENERATE_ARRAY becomes GENERATE_SERIES, DATETIME_ADD becomes `+ INTERVAL '1' HOUR`). so as written this wouldn't run on postgres.

separately it changes `DROP TABLE IF EXISTS mimiciv_derived.icustay_hourly` to unqualified `icustay_hourly`, which would build the table outside the schema every other concept references.

and i think the original problem is already gone: main's generated copy of this file uses `endtime + CAST(hr_unnested AS BIGINT) * INTERVAL '1' HOUR` and `GENERATE_SERIES`, so the interval-construction issue this pr set out to fix no longer appears there.

if there's still a real defect in the bigquery source i'd be glad to look, but it'd need to start there and be regenerated. otherwise this may be worth closing as overtaken by the transpiler work. (the old CI failure here is just the python 3.7 runner going away, unrelated.)

Issue #1876, opened 2025-04-03
Error in mapping of ICD-10 Codes for diabetes  uncomplicated vs diabetes with complications


dewittpe:
Looking at Table 2 in [Quan (2005)](https://journals.lww.com/lww-medicalcare/fulltext/2005/11000/Referral_of_Breast_Cancer_Patients_to_Medical.10.aspx) I see the codes for diabetes, uncomplicated, to be:
>E10.0, E10.1, E10.9,
>E11.0, E11.1, E11.9,
>E12.0, E12.1, E12.9,
>E13.0, E13.1, E13.9,
>E14.0, E14.1, E14.9

And for Diabetes, complicated to be:
>E10.2–E10.8, E11.2–E11.8, E12.2–E12.8, E13.2–E13.8, E14.2–E14.8

From that reading I would expect that codes E106 would be in the complicated section.  

However, in the codes for uncomplicated, the *.6x codes are listed and are omitted from the complected set.

https://github.com/MIT-LCP/mimic-code/blob/679037f9bc12f12995f9a8b304bd169e4b092f10/mimic-iv/concepts/comorbidity/charlson.sql#L170-L193

https://github.com/MIT-LCP/mimic-code/blob/679037f9bc12f12995f9a8b304bd169e4b092f10/mimic-iv/concepts/comorbidity/charlson.sql#L197-L218

alistairewj:
This query is for Charlson comorbidities, which are listed in Table 1 of that paper. This is from Table 1:

![Image](https://github.com/user-attachments/assets/9481f682-71aa-48c5-bafd-d29a7fe48174)

Diabetes uncomplicated includes E10.8, etc, and so the code seems consistent with that.

However we really should bring in an ICD-10 version of Elixhauser!

dewittpe:
@alistairewj : thank you for the gently pointing out I was looking at the wrong table!  I'll review the code I've been working with.

PR #1877, opened 2025-04-03
Fix 1876

dewittpe:
A patch to fix Issue #1876 by mapping the ICD-10 codes for E1[0-4]\.[6,8] which where mapping to Diabetes without complications to map to Diabetes with complications.  This is change is based on the listed codes in Quan (2005) as noted in the issue.


alistairewj:
Closing as mentioned in #1876 - the code is correct for Charlson comorbidities. Feel free to re-open that issue or open a new one if you spot an error still.

PR #1900, opened 2025-05-17
Remove the SQL files that were placed in the wrong folders

yongfanbeta:
Remove the SQL files that were placed in the wrong folders`concepts_postgres`, and add arb.sql in make concepts


alistairewj:
thanks for this - while fixing the actions I stumbled upon it myself (wanted our CI working before I went through PRs - wrong order!)

Closing but the issue is fixed in latest commit!

PR #2042, opened 2026-07-12
fix(mimic-iv): add missing temperature itemids to vitalsign concept

### LINKED ISSUE #1358: Temperature vital sign not consistently captured by concept
There are a few cases of missing temperature `itemid` in the [vitalsign](https://github.com/MIT-LCP/mimic-code/blob/main/mimic-iv/concepts/measurement/vitalsign.sql) derived table - likely they should be added:

* 226329 Blood Temperature CCO (C) 
* When an arctic sun is in use
    * 227632 | Arctic Sun/Alsius Temp # 1 C
    * 227634 | Arctic Sun/Alsius Temp # 2 C
    * 227630 | Arctic Sun Temp # 1 Location
    * 227631 | Arctic Sun Temp # 2 Location

acnimma:
## Summary
- Add missing temperature **itemids** to `vitalsign.sql` (Celsius CASE + site)

### Added itemids (for reviewers)

| itemid | Label | Used as |
| --- | --- | --- |
| **226329** | Blood Temperature CCO (C) | temperature (°C) |
| **227632** | Arctic Sun/Alsius Temp #1 C | temperature (°C) |
| **227634** | Arctic Sun/Alsius Temp #2 C | temperature (°C) |
| **227630** | Arctic Sun Temp #1 Location | `temperature_site` |
| **227631** | Arctic Sun Temp #2 Location | `temperature_site` |

Existing `223762` / `223761` (C/F) and `224642` (site) are unchanged.

## Why
The vitalsign concept omitted blood/CCO and Arctic Sun/Alsius temperature itemids documented in #1358.

## Test plan
- [x] `pytest tests/test_transpile.py` passes locally
- [x] Itemids listed above present in BigQuery + postgres + duckdb copies

Fixes #1358

### DIFF
--- a/mimic-iv/concepts/measurement/vitalsign.sql
+++ b/mimic-iv/concepts/measurement/vitalsign.sql
@@ -53,12 +53,13 @@ SELECT
                     AND valuenum < 120
                     THEN (valuenum - 32) / 1.8
                 -- already in degC, no conversion necessary
-                WHEN itemid IN (223762)
+                -- includes core/blood temps (CCO) and Arctic Sun/Alsius temps
+                WHEN itemid IN (223762, 226329, 227632, 227634)
                     AND valuenum > 10
                     AND valuenum < 50
                     THEN valuenum END)
             AS NUMERIC), 2) AS temperature
-    , MAX(CASE WHEN itemid = 224642 THEN value END
+    , MAX(CASE WHEN itemid IN (224642, 227630, 227631) THEN value END
     ) AS temperature_site
     , AVG(CASE WHEN itemid IN (220277)
                 AND valuenum > 0
@@ -91,10 +92,14 @@ WHERE ce.stay_id IS NOT NULL
         , 220621 -- Glucose (serum)
         , 226537 -- Glucose (whole blood)
         -- TEMPERATURE
-        -- 226329 -- Blood Temperature CCO (C)
+        , 226329 -- Blood Temperature CCO (C)
+        , 227632 -- Arctic Sun/Alsius Temp #1 C
+        , 227634 -- Arctic Sun/Alsius Temp #2 C
         , 223762 -- "Temperature Celsius"
         , 223761  -- "Temperature Fahrenheit"
         , 224642 -- Temperature Site
+        , 227630 -- Arctic Sun Temp #1 Location
+        , 227631 -- Arctic Sun Temp #2 Location
     )
 GROUP BY ce.subject_id, ce.stay_id, ce.charttime
 ;
--- a/mimic-iv/concepts_duckdb/measurement/vitalsign.sql
+++ b/mimic-iv/concepts_duckdb/measurement/vitalsign.sql
@@ -41,13 +41,13 @@ SELECT
         THEN (
           valuenum - 32
         ) / 1.8
-        WHEN itemid IN (223762) AND valuenum > 10 AND valuenum < 50
+        WHEN itemid IN (223762, 226329, 227632, 227634) AND valuenum > 10 AND valuenum < 50
         THEN valuenum
       END
     ) AS DECIMAL(38, 9)),
     2
   ) AS temperature,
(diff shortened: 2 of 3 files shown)


Chessing234:
missing temp itemids are easy to miss. can you list the added itemids in the pr body for reviewers?

acnimma:
Listed the added itemids in the PR body for reviewers:

- Temperature (°C): `226329`, `227632`, `227634`
- `temperature_site`: `227630`, `227631`

All five are in the BigQuery source and the regenerated postgres/duckdb copies.

Chessing234 (changes requested):
the two site itemids look wrong to me. 227630/227631 are arctic sun probe *locations* for the cooling device, so folding them into `MAX(CASE WHEN itemid IN (224642, 227630, 227631) THEN value END) AS temperature_site` puts two different meanings in one column. at a charttime where both are charted the MAX just picks whichever string sorts higher, so temperature_site can end up naming a device probe rather than the site of the 223761/223762 reading sitting next to it. that also silently changes what #2046's `LOWER(temperature_site) LIKE '%axillary%'` matches, and that's your PR too. could you keep temperature_site on 224642 and give the device probes their own column?

226329 was deliberately commented out on main, so uncommenting it needs a reason in the description. blood/CCO is a continuous core temperature on a small cardiac subset, and averaging it into the same per-charttime AVG as oral/axillary moves vitalsign.temperature for those stays. that propagates through first_day_vitalsign.temperature_min/max into apsiii, sofa, sapsii, oasis and sirs. can you show the demo numbers — how many stays change and how far min/max move?

separately the postgres copy doesn't reproduce. regenerating from this branch's bigquery source with the pinned sqlglot==30.11.0 from requirements-lock.txt gives a 6-line diff against what's committed — the header comment block is three lines here and one line from the pinned transpiler. main reproduces byte-identically and the duckdb copy matches, so it looks like it was generated with a different sqlglot.

Chessing234 (changes requested):
thanks for listing the itemids — that made this much easier to check. a few things came out of it.

first the mechanical one: the postgres copy isn't transpiler output. running the repo's transpiler (sqlglot 30.11.0 per requirements-lock) over your bigquery source and diffing against `mimic-iv/concepts_postgres/measurement/vitalsign.sql` on this branch, sqlglot emits the header comment as one line and the branch has it split across three. duckdb matches exactly; the same check against main is identical, so the drift is from this branch. that folder is checked by the `generated-up-to-date` job and the file header says not to edit directly.

the substantive ones:

`226329` was not missing — it's present on main at line 94 of the bigquery source, deliberately commented out (`-- 226329 -- Blood Temperature CCO (C)`). uncommenting an exclusion someone made on purpose needs a reason in the pr body; blood temperature off a CCO/PA catheter is a different measurement site and modality from a nursing-charted temperature, which is presumably why it was left out.

merging the probes into one column via `MAX(...)` worries me more. previously `temperature` had a single source itemid so there was no conflict to resolve. a patient on an Arctic Sun can now have 223762, 227632 and 227634 at the same charttime, and `MAX` silently reports the warmest. for targeted temperature management that's the opposite of what you'd want — the whole point is tracking the cooling probe. same question for `temperature_site`: 224642 is "Temperature Site" but 227630/227631 are "Arctic Sun Temp #1/#2 Location", and `MAX` over text just picks whichever sorts last, so a charttime with both a nursing site and a probe location returns one arbitrarily.

could you show the counts — how many charttimes end up with more than one of these temperature itemids populated, and how many with both a 224642 value and a 227630/227631 value? if that's common then i think these need to be separate columns rather than folded into `temperature`/`temperature_site`.

alistairewj:
These are not included in temperature on purpose. Arctic sun is therapeutic hyperthermia and it's not a measurement of the individual. CCO blood temperature isn't what people think of when they think of temperature - maybe it's a legitimate measurement but I think that would need to be demonstrated rather than assumed.

PR #2043, opened 2026-07-12
fix(mimic-iv): update Charlson ICD-10-CM codes for C4A, C7A, and C7B

### LINKED ISSUE #2017: ICD-10-CM codes captured and not captured in charlson.sql
ICD-10-CM have three categories which are not part of the WHO ICD-10 and as such are not defined in Quan et.al (2005)

* C4A: Merkel cell carcinoma
* C7A: Malignant neuroendocrine tumors
* C7B: Secondary neuroendocrine tumors

One of the lines in the SQL for mapping ICD-10 to malignant cancer has a possible positive unexpected side effect

https://github.com/MIT-LCP/mimic-code/blob/57069783095e7770e66ea97da264c0200078ddbf/mimic-iv/concepts/comorbidity/charlson.sql#L305

This will map C4A codes to malignant cancer.   That seems conceptually okay but it opens the door for what to do with C7A and C7B codes.

If C4A is included in malignant cancer then shouldn't C7A codes?

Should C4A be excluded to be consistent with Quan et.al (2005)?

I think Merkel cells are part of the skin and since Quan et.al (2005) defines "any malignancy, including lymphoma and leukemia, except malignant neoplasm of skin" that C4A should be excluded?

Cancer is not my area of expertise, I honest have no idea.  This is more of question of being consistent with the publication.

acnimma:
## Summary
- Exclude C4A, include C7A in `malignant_cancer`, C7B in `metastatic_solid_tumor`
- Regenerate postgres and duckdb copies from BigQuery source

## Test plan
- [x] `pytest tests/test_transpile.py` (161 tests)
- [x] Transpiled postgres/duckdb `charlson.sql`

Fixes #2017

### DIFF
--- a/mimic-iv/concepts/comorbidity/charlson.sql
+++ b/mimic-iv/concepts/comorbidity/charlson.sql
@@ -285,6 +285,9 @@ WITH diag AS (
 
         -- Any malignancy, including lymphoma and leukemia,
         -- except malignant neoplasm of skin.
+        -- ICD-10-CM codes added after Quan et al. (2005):
+        --   C4A (Merkel cell carcinoma) excluded as skin malignancy per Quan.
+        --   C7A (malignant neuroendocrine tumors) included as primary malignancy.
         , MAX(CASE WHEN
             SUBSTR(icd9_code, 1, 3) BETWEEN '140' AND '172'
             OR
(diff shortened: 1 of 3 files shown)


Chessing234:
c4a/c7a/c7b handling is the right call for quan 2005. worth a short comment in the sql so the next person doesn't "fix" it back in.

acnimma:
Quan notes in charlson.sql thanks

Chessing234 (changes requested):
i checked the code ranges rather than taking them on trust, and the two additions do hold up: 'C7A' sorts above 'C76' and below 'C81', so it genuinely falls outside every existing range, and 'C7B' isn't in the `IN ('C77','C78','C79','C80')` list. the C4A exclusion is also doing real work — 'C4A' sorts between 'C45' and 'C58', so without it Merkel cell would have been picked up by that range. so the logic is sound and the operator precedence works out the way you intended.

two things before this lands.

the postgres copy isn't transpiler output. running the repo's transpiler (sqlglot 30.11.0 from requirements-lock) against your bigquery source and diffing `mimic-iv/concepts_postgres/comorbidity/charlson.sql` on this branch, sqlglot emits the reference header as a single line and the branch has it wrapped across 16. the duckdb copy matches exactly, and the same check on main is identical, so this branch introduces it. `generated-up-to-date` in transpile.yml compares that folder against transpiler output.

the one i'd actually push back on: excluding C4A while C43 stays in is inconsistent. your comment says C4A is excluded "as skin malignancy per Quan", but Quan's ICD-10 list explicitly includes C43 (malignant melanoma of skin) — main already has it in the `IN ('C43','C88')` branch. the Charlson convention is to exclude non-melanoma skin cancer (C44) and count melanoma. Merkel cell is an aggressive cutaneous malignancy, much closer to melanoma than to C44, so counting C43 but not C4A is hard to defend on the rationale given. i'd either drop the C4A exclusion or explain why it's grouped with C44 rather than C43.

and the evidence question: Quan 2005 predates all three codes, so this is a deliberate extension of the published algorithm rather than a correction to it — worth saying so in the body. could you also show how many hadm_ids actually change? C4A/C7A/C7B entered ICD-10-CM well into MIMIC-IV's coverage window, and if they appear in `diagnoses_icd` a handful of times or not at all, that's worth knowing before changing a widely-used comorbidity concept.

PR #2059, opened 2026-07-16
FLOOR postgres DATETIME_DIFF to match BigQuery

### LINKED ISSUE #1549: `DATETIME_DIFF` implementation varies between postgres and bigquery
As raised in #1529, `DATETIME_DIFF` in PostgreSQL calculates a fractional number, whereas `DATETIME_DIFF` in BigQuery returns a whole number. This results in a subtle divergence of the concepts in the two dialects. urine_output_rate may be affected as various comparisons expected fractional values to be returned:

https://github.com/MIT-LCP/mimic-code/blob/a8308706d8f2bf3bbf42f5e7065094c648f64576/mimic-iv/concepts/measurement/urine_output_rate.sql#L44-L53

Easiest fix is to (1) update postgres-functions.sql and (2) update the UO query to take a similar approach to the kdigo_uo query.

Chessing234:
## Summary
- Postgres `DATETIME_DIFF` returned fractional epochs; BigQuery truncates (#1549)
- Wrap ratio in `FLOOR` in iii/iv `postgres-functions.sql`

## Test plan
- [ ] `DATETIME_DIFF` minute/hour diffs match BQ for same timestamps
- [ ] Existing concept queries still run

Fixes #1549

### DIFF
--- a/mimic-iii/concepts_postgres/postgres-functions.sql
+++ b/mimic-iii/concepts_postgres/postgres-functions.sql
@@ -71,7 +71,8 @@ LANGUAGE PLPGSQL;
 -- i.e. encapsulate it in single quotes
 CREATE OR REPLACE FUNCTION DATETIME_DIFF(endtime TIMESTAMP(3), starttime TIMESTAMP(3), datepart TEXT) RETURNS NUMERIC AS $$
 BEGIN
-RETURN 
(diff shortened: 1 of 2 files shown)


alistairewj:
these functions are obsolete and slated for removal!

Chessing234:
ah fair enough, thanks for the heads up

PR #2071, opened 2026-07-19
fix(mimic-iv): retain device-only rows in oxygen_delivery

### LINKED ISSUE #1570: Question about the number of patients in the oxygen_delivery table
### Prerequisites

### Description

Hello, 

We have observed that certain patients exhibit results in the chartevents table with itemid = 226732, but these results are absent in the oxygen_delivery table. One example is stay_id = 30813912.

Upon analyzing line 86 of the oxygen_delivery table, we discovered that placing the condition "where ce.rn = 1" here only includes patients present in the ce_stg2 table. Consequently, data for patients existing in the o2 table but not in the ce_stg2 table is missing.

Our solution is to remove this line of code and create a new table, ce_stg3. Do you have any suggestions for a better solution? 

Thank you!

Daonan Chen & Zhou Zhou

`, ce_stg2 AS
(
  select
    ce.subject_id
    , ce.stay_id
    , ce.charttime
    , itemid
    , value
    , valuenum
    , valueuom
    -- retain only 1 row per charttime
    -- prioritizing the last documented value
    -- primarily used to subselect o2 flows
    , ROW_NUMBER() OVER (PARTITION BY subject_id, charttime, itemid ORDER BY storetime DESC) as rn
  FROM ce_stg1 ce
)
, ce_stg3 AS
(
	SELECT * FROM ce_stg2 WHERE rn = 1
)`

acnimma:
## Summary
- Filter O2 flow `rn` in `ce_stg3` before the FULL OUTER JOIN in `oxygen_delivery.sql`
- Retain device-only charttimes (itemid 226732) that previously disappeared when `WHERE ce.rn = 1` ran after the join

## Test plan
- [x] `pytest tests/test_transpile.py -k oxygen_delivery` passes locally

`oxygen_delivery` FULL OUTER JOINs O2 flow (`ce_stg2`) with delivery devices (`o2`, itemid 226732), then applied `WHERE ce.rn = 1`. That filter drops rows where `ce` is NULL, so stays with device charting but no flow (e.g. `stay_id = 30813912` in #1570) never appear in the derived table. Move the `rn = 1` filter into `ce_stg3` before the join so device-only rows are kept. Regenerate postgres and duckdb dialect copies from the BigQuery source.

Fixes #1570

### DIFF
--- a/mimic-iv/concepts/measurement/oxygen_delivery.sql
+++ b/mimic-iv/concepts/measurement/oxygen_delivery.sql
@@ -47,6 +47,21 @@ WITH ce_stg1 AS (
     FROM ce_stg1 ce
 )
 
+-- filter rn before joining so device-only rows from the FULL OUTER JOIN
+-- are retained (a post-join WHERE ce.rn = 1 would drop rows where ce is NULL)
+, ce_stg3 AS (
+    SELECT
+        subject_id
+        , stay_id
+        , charttime
+        , itemid
+        , value
+        , valuenum
+        , valueuom
+    FROM ce_stg2
+    WHERE rn = 1
+)
+
 , o2 AS (
     -- The below ITEMID can have multiple entries for charttime/storetime
     -- These are valid entries, and should be retained in derived tables.
@@ -85,12 +100,10 @@ WITH ce_stg1 AS (
         , ce.valuenum
         , o2.o2_device
         , o2.rn
-    FROM ce_stg2 ce
+    FROM ce_stg3 ce
     FULL OUTER JOIN o2
-                    ON ce.subject_id = o2.subject_id
-        AND ce.charttime = o2.charttime
-    -- limit to 1 row per subject_id/charttime/itemid from ce_stg2
-    WHERE ce.rn = 1
+        ON ce.subject_id = o2.subject_id
+            AND ce.charttime = o2.charttime
 )
 
 SELECT
--- a/mimic-iv/concepts_duckdb/measurement/oxygen_delivery.sql
+++ b/mimic-iv/concepts_duckdb/measurement/oxygen_delivery.sql
@@ -24,6 +24,18 @@ WITH ce_stg1 AS (
     valueuom,
     ROW_NUMBER() OVER (PARTITION BY subject_id, charttime, itemid ORDER BY storetime DESC, valuenum DESC) AS rn
   FROM ce_stg1 AS ce
+), ce_stg3 AS (
+  SELECT
+    subject_id,
+    stay_id,
+    charttime,
+    itemid,
+    value,
+    valuenum,
+    valueuom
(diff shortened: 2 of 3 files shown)


Chessing234 (commented):
the post-join `where ce.rn = 1` was definitely nuking device-only rows from the full outer join — moving the rn filter into ce_stg3 first is the right shape. coalesce on subject/stay/charttime was already there so device-only keys should survive. if you haven't already, worth a quick count of rows where o2_flow is null but o2_delivery_device_1 isn't, before vs after.

Chessing234:
keeping device-only oxygen rows is useful for vent studies. does the regenerated postgres/duckdb copy land in this pr too?

Chessing234 (commented):
filtering rn before the full outer join is the right fix — post-join `where ce.rn = 1` was dropping device-only o2 rows when ce was null. mirrored across dialects; the comment in the bq file helps. a tiny regression query in concepts/documentation (device-only count before/after) would make the behavioral change obvious to reviewers.

acnimma:
dialects included; skip regression query (nice-to-have, not required thank you

Chessing234 (changes requested):
following up on the dialect question above — the postgres copy in this pr isn't what the transpiler produces.

i ran the repo's own transpiler (sqlglot 30.11.0 from requirements-lock) over your bigquery source and diffed it against `mimic-iv/concepts_postgres/measurement/oxygen_delivery.sql` as committed here. the `o2` comment block comes out of sqlglot as a single line, but this pr has it re-wrapped across 12 separate `/* ... */` lines. the duckdb copy is byte-identical, so it looks like just the postgres file was touched by hand. running the same check against main gives IDENTICAL for this file, so the drift starts in this branch.

that file's header says it's automatically generated and not to be edited directly, and the `generated-up-to-date` job in transpile.yml diffs it against transpiler output — so this will come back as a failure once the checks actually run (the current ones were cancelled waiting for runners, not failed).

could you regenerate the postgres copy instead of editing it? the `ce_stg3` logic itself looks right to me — moving `rn = 1` ahead of the full outer join does retain the device-only rows, and the `COALESCE` on subject_id/charttime in `stg` means those rows keep their grouping keys.

one thing i'm not sure about and would want your read on: with device-only charttimes now retained, those rows land in the output with `o2_flow` null. worth confirming that's the intended shape for #1570 rather than something downstream consumers would need to filter.

alistairewj (approved):

alistairewj:
It was intentional to exclude device only rows - e.g. the device only row in the given example is `Endotracheal Tube` - this table was mostly meant to capture oxygen delivery with other devices. It *is* a bit misleading that o2 device isn't present in this table but if we merge this change as is, we get an overwhelming number of rows with empty o2 flow and an o2 device of ETT. I think a better fix is a separate table with oxygen_device

PR #2072, opened 2026-07-19
fix: quote datadir for postgres load and wget

Chessing234:
## Summary
- load/wget still broke on paths with spaces after #2068
- quote `DATADIR` for psql `-v` / wget `-P`, and `\cd :'mimic_data_dir'`

## Test plan
- [ ] `make mimic-build-gz datadir="/tmp/foo bar/"` with csv.gz present loads
- [ ] path without spaces still works


Chessing234:

PR #2076, opened 2026-07-19
fix: quote sqlite import.sh paths for spaces

Chessing234:
## Summary
- `.import` / `gzip -dc` broke when the csv path had spaces
- quote paths in iii + iv `import.sh`

## Test plan
- [ ] import a csv from a directory with a space in the name


Chessing234:

PR #2079, opened 2026-07-19
fix: use DATETIME_DIFF in neuroblock_dose for BigQuery

Chessing234:
## Summary
- BQ concepts still used postgres `interval` subtraction for the 8h gap check
- switch to `DATETIME_DIFF(..., HOUR)` like the other duration scripts

## Test plan
- [ ] BigQuery dry-run of neuroblock_dose.sql


Chessing234:
folded into #2124

PR #2082, opened 2026-07-19
fix: load plain .csv in mimic-iii sqlite import.py

Chessing234:
## Summary
- import.py only globbed `*.csv.gz`; plain `.csv` was skipped
- prefer `.csv.gz` when both exist for the same table

## Test plan
- [ ] run import.py with only `.csv` files present


Chessing234:

PR #2083, opened 2026-07-19
fix: use printf for duckdb load progress (no literal \c)

Chessing234:
## Summary
- `echo ... \c` prints a literal `\c` on many shells
- use `printf` so "done!" stays on the same line

## Test plan
- [ ] run duckdb import; progress lines look normal


Chessing234:

PR #2085, opened 2026-07-19
fix: low_memory=False on mimic-iv sqlite read_csv

Chessing234:
## Summary
- non-chunked and patients preview reads omitted `low_memory=False`
- matches the chunked path and avoids mixed-dtype warnings

## Test plan
- [ ] import.py on demo data; no dtype warnings


Chessing234:

PR #2086, opened 2026-07-19
fix: escape quotes in duckdb COPY file paths

Chessing234:
## Summary
- paths with `'` broke the COPY ... FROM '...' string
- escape by doubling quotes (iii/iv/ed/note)

## Test plan
- [ ] duckdb import with a quoted path segment succeeds


Chessing234:

PR #2087, opened 2026-07-19
fix: quote DIRNAME in duckdb import case

Chessing234:
## Summary
- unquoted `case $DIRNAME` breaks when the folder name has spaces
- quote it in iv/ed/note duckdb import scripts

## Test plan
- [ ] import with a datadir whose leaf folder has a space


Chessing234:

PR #2089, opened 2026-07-19
fix: use globs instead of ls in make_concepts.sh

Chessing234:
## Summary
- replace `ls $d` with `"$d"/*.sql` so filenames with spaces don't break
- quote destination table and sql file paths on bq query redirects

## Test plan
- [ ] `bash -n mimic-iv/concepts/make_concepts.sh`


Chessing234:

PR #2090, opened 2026-07-19
fix: use globs and quote paths in cxr label scripts

Chessing234:
## Summary
- replace `ls` loops with globs in negbio/chexpert runners
- quote paths so folders with spaces don't break

## Test plan
- [ ] `bash -n` on both scripts


Chessing234:

PR #2091, opened 2026-07-19
fix: check $1 not pid in copy_concepts script

Chessing234:
## Summary
- `$$1` was the shell pid, so the usage check never fired
- quote bq dataset/table args

## Test plan
- [ ] `bash -n mimic-iv/concepts/copy_concepts_to_versioned_schema.sh`
- [ ] running with no args prints usage and exits 1


Chessing234:
folded into #2124

PR #2092, opened 2026-07-19
fix: write transpiled sql as utf-8

Chessing234:
## Summary
- open destination files with encoding=utf-8 (reads already use utf-8-sig)

## Test plan
- [ ] transpile smoke / existing transpile unit tests


Chessing234:

PR #2093, opened 2026-07-19
fix: match validate_concepts tables by exact name

Chessing234:
## Summary
- use awk field equality instead of regex grep so short table names don't false-match
- use `grep -Fx` when checking expected set membership

## Test plan
- [ ] `bash -n mimic-iv/concepts/validate_concepts.sh`


alistairewj:
is this actually a bug? the stated comment is incorrect, the grep can't hit prefixes it has a trailing `,`

Chessing234:
yeah fair — the trailing comma already stops prefix matches. i overstated that. closing.

PR #2080, opened 2026-07-19
fix: duckdb iii usage comment says mimic3.db

Chessing234:
## Summary
- comment said mimic4.db; default outfile is mimic3.db

## Test plan
- [ ] n/a (comment only)

PR #2094, opened 2026-07-19
fix: duckdb error text cites import_duckdb.sh

Chessing234:
## Summary
- error text said `import.sh` instead of `import_duckdb.sh` in iii/iv/ed/note duckdb importers

## Test plan
- [ ] `bash -n` on the four import_duckdb.sh scripts


Chessing234:
default db comment is already covered in #2080 — this one's just the error text.

alistairewj (approved):

PR #2109, opened 2026-07-19
fix(mimic-iv-note): correct postgres quickstart data paths

acnimma:
## Summary
- `mimic_data_dir` -> `mimic-iv/2.2/note` after the wget `mv`
- Fix cleanup paths and remove stray `rmdir rm`

## Test plan
- [ ] Paths match a fresh wget + mv tree

The note postgres quickstart renames the download tree to `mimic-iv`, but still pointed `mimic_data_dir` at the nested `mimic-iv/mimic-iv-note/...` path and used `rmdir rm` in the cleanup comment.


alistairewj:
the commented out paths are still incorrect but I'm merging for convenience since they are commented out and the real fix is there

PR #2110, opened 2026-07-19
fix(mimic-iv-ed): quote postgres mimic_data_dir for paths with spaces

acnimma:
## Summary
- Use `\cd :''mimic_data_dir''` in ED `load.sql` / `load_gz.sql` / `load_7z.sql`

## Test plan
- [ ] Load with `mimic_data_dir` containing a space succeeds
- [ ] Path without spaces still works

Same space-safe psql quoting as the III datadir fix: unquoted `\cd :mimic_data_dir` splits on spaces. Quote the variable in all three ED load scripts.


alistairewj:
same as #2111 and #2116 this broke the build on my nix machine so please add CI to verify it is fixing a real issue

PR #2111, opened 2026-07-19
fix(mimic-iv): quote mimic_data_dir in uncompressed load.sql

acnimma:
## Summary
- Use `\cd :''mimic_data_dir''` in `mimic-iv/.../load.sql`

## Test plan
- [ ] Uncompressed load with a spaced datadir path succeeds

The uncompressed IV `load.sql` still used unquoted `\cd :mimic_data_dir`. Apply the same psql quoted-variable form used for space-safe datadir handling.


alistairewj:
this breaks the build I'm pretty sure - the meta command can handle spaces.

PR #2115, opened 2026-07-19
fix(mimic-iv-note): quote postgres mimic_data_dir for paths with spaces

acnimma:
## Summary
- Use `\cd :''mimic_data_dir''` in note `load.sql` / `load_gz.sql` / `load_7z.sql`

## Test plan
- [ ] Load with a spaced datadir path succeeds
- [ ] Path without spaces still works

Same space-safe psql quoting as III/ED: unquoted `\cd :mimic_data_dir` splits on spaces.


alistairewj:
similar to #2111 #2116 and #2110 if you are making AI pull requests please try not to split the same fix across 4 PRs it's difficult to manage

PR #2116, opened 2026-07-19
fix(mimic-iv): quote mimic_data_dir in load_gz and load_7z

acnimma:
## Summary
- Use `\cd :''mimic_data_dir''` in `load_gz.sql` and `load_7z.sql`

## Test plan
- [ ] Compressed load with a spaced datadir path succeeds

Companion to the uncompressed `load.sql` quoting fix: gz/7z loaders still used unquoted `\cd :mimic_data_dir`.


alistairewj:
same as this one - pretty sure this breaks the build - if you are modifying this code for robustness please add a CI case and verify it actually works before the PR

PR #2118, opened 2026-07-19
fix: quote paths in mimic-iii make-concepts.sh

Chessing234:
## Summary
- quote bq destination tables and sql/csv paths so spaces work

## Test plan
- [ ] dry-run make-concepts from a path with spaces


Chessing234:

PR #2120, opened 2026-07-20
fix(mimic-iv): do not treat CANCELLED cultures as positive

### LINKED ISSUE #1677: identify negative culture
The org_name column in the microbiologyevents table contains the data "CANCELLED" (org_itemid: 90760). Is this a case  the culture was implemented but the culture was cancelled?

In order to identify negative culture, should I add the condition if org_name="CANCELLED"?

https://github.com/MIT-LCP/mimic-code/blob/main/mimic-iv/concepts_postgres/sepsis/suspicion_of_infection.sql

```
me AS (
    SELECT micro_specimen_id
        -- the following columns are identical for all rows
        -- of the same micro_specimen_id
        -- these aggregates simply collapse duplicates down to 1 row
        , MAX(subject_id) AS subject_id
        , MAX(hadm_id) AS hadm_id
        , CAST(MAX(chartdate) AS DATE) AS chartdate
        , MAX(charttime) AS charttime
        , MAX(spec_type_desc) AS spec_type_desc
        -- identify negative cultures as NULL organism
        -- or a specific itemid saying "NEGATIVE"
        , MAX(
            CASE WHEN org_name IS NOT NULL
                      AND org_itemid != 90856
                      AND org_name != ''
                THEN 1 ELSE 0
            END) AS positiveculture
    FROM mimiciv_hosp.microbiologyevents
    GROUP BY micro_specimen_id
)
```

acnimma:
﻿## Summary
- Exclude `org_itemid` 90760 (`CANCELLED`) from `positiveculture` in `suspicion_of_infection.sql`
- Also exclude `org_name = 'CANCELLED'`
- Update postgres and duckdb dialect copies

## Test plan
- [x] Source + dialect copies exclude 90760 / CANCELLED
- [ ] Specimen with only CANCELLED → `positive_culture = 0`
- [ ] Specimen with a real organism → `positive_culture = 1`

`positiveculture` treated any non-NULL, non-empty `org_name` as positive except NEGATIVE (`90856`). CANCELLED cultures (`org_itemid` 90760) are not microbial growth, but were counted as positive and could inflate `positive_culture` on the suspicion-of-infection table. Exclude CANCELLED the same way NEGATIVE is excluded.

Fixes #1677

### DIFF
--- a/mimic-iv/concepts/sepsis/suspicion_of_infection.sql
+++ b/mimic-iv/concepts/sepsis/suspicion_of_infection.sql
@@ -30,12 +30,13 @@ WITH ab_tbl AS (
         , CAST(MAX(chartdate) AS DATE) AS chartdate
         , MAX(charttime) AS charttime
         , MAX(spec_type_desc) AS spec_type_desc
-        -- identify negative cultures as NULL organism
-        -- or a specific itemid saying "NEGATIVE"
+        -- non-positive: NULL/empty organism, NEGATIVE (90856),
+        -- or CANCELLED culture (90760) — cancelled is not growth
         , MAX(
             CASE WHEN org_name IS NOT NULL
-                      AND org_itemid != 90856
+                      AND org_itemid NOT IN (90856, 90760)
                       AND org_name != ''
+                      AND org_name != 'CANCELLED'
                 THEN 1 ELSE 0
             END) AS positiveculture
     FROM `physionet-data.mimiciv_hosp.microbiologyevents`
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
yeah treating CANCELLED as growth was pretty misleading for soi. excluding 90760 and the org_name string both seems fine (belt and suspenders). only thing i'd double-check is whether any other org_itemids are similarly non-organism (contaminated/etc) — not blocking, just curious.

Chessing234:
excluding cancelled cultures is overdue. worth noting in a comment next to the filter so future edits don't reintroduce them.

Chessing234 (commented):
good catch — cancelled cultures aren't growth. covering itemid 90760 and org_name 'CANCELLED' in all three dialects is the right shape. maybe add a one-line comment near the CASE that 90760 is the cancelled organism itemid so the next reader doesn't have to dig microbiologyevents for it.

acnimma:
present next to CASE thank you

alistairewj (approved):

PR #2121, opened 2026-07-20
fix(mimic-iii): do not bridge paused vasopressor gaps

### LINKED ISSUE #1808: paused vasopressor durations
### Prerequisites

### Description

Hi,

in lines 264 to 271 in https://github.com/MIT-LCP/mimic-code/blob/main/mimic-iii/concepts/durations/vasopressor_durations.sql, vasopressor administrations from `inputevents_mv` are grouped based on their `linkorderid`, and the minimum and maximum `starttime` and `endtime`, respectively, are calculated for each group. This makes sense if the `starttime` of an entry equals the `endtime` of the previous entry, but not if there is a gap between. According to https://github.com/MIT-LCP/mimic-code/issues/47, this should never be the case after removing "Rewritten" entries, but I found at least one counterexample, namely `linkorderid` = `8781494`:

```
select
    icustay_id, linkorderid, starttime, endtime, statusdescription, itemid
  FROM inputevents_mv
  where linkorderid = 8781494
  order by starttime
```

In the result, there is a ~6h gap between `2167-08-05 21:16:00` and `2167-08-06 04:01:00`. To my understanding, such gaps should not be included in the final vasopressor durations.

**Note**: The `status` of the entry before the gap is "Paused". I did not investigate whether "Paused" entries always lead to gaps, but can do that if it helps.

acnimma:
## Summary
- Stop collapsing MetaVision vasopressor rows with `GROUP BY linkorderid` + `min/max` times
- Keep each `inputevents_mv` row as its own interval; `vasomv_grp` still merges contiguous/overlapping intervals
- Update postgres and duckdb dialect copies
- Document the paused-gap rationale in SQL (BigQuery source + postgres copy cite #1808)

## Why
`vasopressor_durations` grouped MetaVision rows by `linkorderid` and took `min(starttime)` / `max(endtime)`. That is correct when successive rows abut, but after `statusdescription = 'Paused'` there is often a multi-hour gap (example `linkorderid = 8781494`, ~6h; see #1808, median ~1h 11m). Bridging that gap overstates vasopressor exposure.

## Duration threshold / behavior
- **Paused gaps:** leave as separate intervals (do not bridge via `linkorderid`)
- **Contiguous / overlapping:** still merged by `vasomv_grp`

## Test plan
- [x] No `GROUP BY linkorderid` remains in `vasomv`
- [x] SQL comments cite paused-gap / #1808 in BigQuery + postgres (duckdb transpile drops comments)
- [ ] `linkorderid = 8781494` (Paused gap ~6h) yields separate durations, not one bridged interval
- [ ] Contiguous non-paused orders still merge via `vasomv_grp`

Fixes #1808

### DIFF
--- a/mimic-iii/concepts/durations/vasopressor_durations.sql
+++ b/mimic-iii/concepts/durations/vasopressor_durations.sql
@@ -40,7 +40,7 @@ with io_cv as
 , io_mv as
 (
   select
-    icustay_id, linkorderid, starttime, endtime
+    icustay_id, starttime, endtime
   FROM `physionet-data.mimiciii_clinical.inputevents_mv` io
   -- Subselect the vasopressor ITEMIDs
   where itemid in
@@ -259,15 +259,17 @@ WHERE NOT EXISTS(SELECT * FROM vasocv s2
 GROUP BY s1.icustay_id, s1.starttime
 ORDER BY s1.icustay_id, s1.starttime
 )
--- now we extract the associated data for metavision patients
--- do not need to group by itemid because we group by linkorderid
+-- MetaVision rows: keep each order interval as-is.
+-- Do not collapse by linkorderid: Paused rows often leave multi-hour gaps
+-- before the next row of the same linkorderid (#1808). vasomv_grp still
+-- merges contiguous/overlapping intervals afterward.
 , vasomv as
 (
   select
-    icustay_id, linkorderid
-    , min(starttime) as starttime, max(endtime) as endtime
+    icustay_id
+    , starttime
+    , endtime
   from io_mv
-  group by icustay_id, linkorderid
 )
 , vasomv_grp as
 (
--- a/mimic-iii/concepts_duckdb/durations/vasopressor_durations.sql
+++ b/mimic-iii/concepts_duckdb/durations/vasopressor_durations.sql
@@ -30,7 +30,6 @@ WITH io_cv AS (
 ), io_mv AS (
   SELECT
     icustay_id,
-    linkorderid,
     starttime,
     endtime
   FROM mimiciii.inputevents_mv AS io
@@ -166,13 +165,9 @@ WITH io_cv AS (
 ), vasomv AS (
   SELECT
     icustay_id,
-    linkorderid,
-    MIN(starttime) AS starttime,
-    MAX(endtime) AS endtime
+    starttime,
+    endtime
   FROM io_mv
-  GROUP BY
-    icustay_id,
-    linkorderid
 ), vasomv_grp AS (
   SELECT
     s1.icustay_id,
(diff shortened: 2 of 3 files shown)


Chessing234 (commented):
nice catch on the linkorderid collapse — that paused gap example (~6h) really shouldn't get glued into one interval. leaving vasomv_grp to merge only contiguous/overlapping bits feels like the right split. maybe worth a one-liner in the test plan for a non-paused contiguous pair still merging, just so nobody regresses that later.

Chessing234:
not bridging paused vaso gaps is the right clinical call. can you cite the duration threshold you used in the pr description?

Chessing234:
keeping paused gaps as separate intervals matches the raw inputevents_mv semantics better than collapsing on linkorderid. can you add a short comment in the sql citing the paused-gap example so future edits dont reintroduce the group by?

Chessing234 (commented):
dropping the linkorderid min/max collapse is the right fix for paused vaso gaps — contiguous merge in vasomv_grp still covers true overlaps. mirrored across dialects which is what this repo needs. a one-line comment in the postgres/duckdb copies pointing at #1808 (like the bq file already has) would help the next reader.

acnimma:
Addressing the review notes:

- **Duration threshold:** documented in the PR body — example paused gap is `linkorderid = 8781494` (~6h); #1808 reports median ~1h 11m. We do not bridge those gaps.
- **SQL comments:** BigQuery source has the paused-gap / `linkorderid` note above `vasomv`; postgres dialect copy carries an inline `/* ... #1808 */` on the `vasomv` CTE (duckdb transpile strips comments by design).
- **Contiguous merge:** still covered by `vasomv_grp`; called out explicitly in the test plan so that path is not regressed.

Chessing234 (changes requested):
two things here, one mechanical and one i'd like your read on.

the postgres copy doesn't match what the transpiler produces. i ran the repo's own transpiler (sqlglot 30.11.0 from requirements-lock) over your bigquery source and diffed against `mimic-iii/concepts_postgres/durations/vasopressor_durations.sql` as committed on this branch: sqlglot renders your four-line comment as a run of separate `/* ... */` fragments on one line, but the branch has a single hand-written `/* keep each MetaVision interval; do not bridge Paused gaps via linkorderid (#1808) */`. the duckdb copy is byte-identical to transpiler output, so it's only the postgres file. running the same check against main returns identical for this file, so the drift starts in this branch. the `generated-up-to-date` job in transpile.yml diffs that folder against transpiler output, and the file header says not to edit it directly.

on the substance — i think not bridging paused gaps is the right call, but as written the query has no threshold at all despite the "duration threshold / behavior" heading. `vasomv_grp` only merges where `s1.starttime <= t1.endtime`, so once the linkorderid grouping is gone, two rows of the same order separated by one second become two separate durations. #1808's median gap of ~1h11m establishes that the paused gaps are large; it doesn't establish that small ones are absent. could you post the distribution of gaps between consecutive rows within a linkorderid — specifically how many fall under a few minutes? if there's a tail of sub-minute charting gaps then this needs a tolerance rather than zero, and that's also the number that would let the two unchecked boxes in your test plan be checked.

separately, worth calling out explicitly: the carevue branch above keeps its own gap handling, so after this change the same output table carries two different definitions of when an interval breaks depending on which system the stay came from. a line in the sql or the pr body would help anyone unioning the two.

PR #2126, opened 2026-07-25
Fix SOFA CV scoring, urine_output_rate diffs, and Elixhauser no_drg

Chessing234:
## Summary
- SOFA CV score 3 now requires epi/norepi rate > 0 (rate 0 no longer scores 3)
- `urine_output_rate` uses second-based DATETIME_DIFF so BigQuery and Postgres agree (#1549)
- Elixhauser AHRQ no_drg excludes primary diagnoses as documented and RTRIMs ICD-9 codes (#1077)

## Test plan
- [x] `pytest tests/` for changed concepts (sqlfluff + transpile)
- [x] Re-transpiled postgres/duckdb dialects for edited BigQuery sources
- [ ] CI transpile / concepts workflows

Made with [Cursor](https://cursor.com)

PR #2124, opened 2026-07-23
fix: load paths, imports, transpile, and CI permissions

Chessing234:
## Summary
- Quote/escape data dirs and passwords in postgres/sqlite/duckdb loaders
- Accept plain `.csv` (and `.csv.gz`) in duckdb/sqlite imports; utf-8 transpile writes
- sqlfluff annotate no longer fails fork PRs on missing permissions
- Concept SQL changes removed — landing in separate PRs (#2126 and follow-ups)

## Test plan
- [x] `pytest tests/test_transpile.py`
- [ ] smoke `import_duckdb.sh` / sqlite import with a spaced datadir

PR #2127, opened 2026-07-25
fix: load paths, imports, transpile, and CI permissions

Chessing234:
Replacement for #2124 after dropping concept SQL (now in #2126 / follow-ups).


Chessing234:
Superseded by #2128 (clean rebase onto main — same infra changes, no stale concept three-dot noise).

PR #2137, opened 2026-07-29
fix: compare APS III equidistant arms to min not max

Chessing234:
## Summary
- APS III “worst value” ties for resp rate, hematocrit, WBC, sodium, albumin, and glucose compared `ABS(max − mid) = ABS(max − mid)` instead of max vs min.
- Align those arms with heart rate / MBP / temperature (and Knaus APS III furthest-from-normal selection). Updated MIMIC-III and MIMIC-IV BigQuery, Postgres, and DuckDB copies.

## Test plan
- [ ] No remaining `ABS(*_max*) = ABS(*_max*)` in any `apsiii.sql`
- [ ] Equidistant example: glucose min=60, max=200 → still picks larger component score
- [ ] sqlfluff / transpile CI on changed concepts


Chessing234:

PR #2138, opened 2026-07-29
fix: cap MIMIC-III MELD initial at raw sum > 4

Chessing234:
## Summary
- MIMIC-III `meld.sql` capped when the pre-scale creatinine+bilirubin+INR sum was `> 40`, so the OPTN max of 40 almost never applied (raw sum is typically ~1–5 before `round(...)*10`).
- Change the threshold to `> 4` to match MIMIC-IV and [OPTN MELD policy](https://optn.transplant.hrsa.gov/news/meld-serum-sodium-policy-changes/) (max 40 after rounding to one decimal and ×10). Updated BigQuery, Postgres, and DuckDB copies.

## Test plan
- [ ] Spot-check: raw sum `4.1` → `meld_initial` 40 (was 41 after round×10)
- [ ] Spot-check: raw sum `3.95` → `meld_initial` 40 via round×10
- [ ] Diff threshold against `mimic-iv/concepts/organfailure/meld.sql`

### DIFF
--- a/mimic-iii/concepts/organfailure/meld.sql
+++ b/mimic-iii/concepts/organfailure/meld.sql
@@ -124,7 +124,7 @@ left join `physionet-data.mimiciii_derived.rrt_first_day` r
     , inr_score
 
     , case
-        when (creatinine_score + bilirubin_score + inr_score) > 40
+        when (creatinine_score + bilirubin_score + inr_score) > 4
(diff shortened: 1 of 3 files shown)


Chessing234:

alistairewj (approved):

PR #2139, opened 2026-07-29
fix: parse-safe D/C match in pivoted_invasive_lines

Chessing234:
## Summary
- BigQuery `pivoted_invasive_lines` used `'D/C\'d'`, which sqlglot cannot parse (same class of failure as neuroblock).
- Switch to `stopped LIKE 'D/C%'` and regenerate postgres/duckdb copies.

## Test plan
- [x] `pytest tests/test_transpile.py -k invasive` (4 passed)
- [ ] Confirm transpile of this concept succeeds in CI


Chessing234:

PR #2146, opened 2026-07-31
fix: include wbc_max in SIRS WBC missing-data guard

Chessing234:
## Summary
- SIRS WBC scoring uses `wbc_min`, `wbc_max`, and `bands_max`, but the missing-data arm only checked `COALESCE(wbc_min, bands_max)`.
- When only `wbc_max` is present (and normal), the CASE fell through to NULL instead of 0.
- Update BigQuery, Postgres, and DuckDB copies for MIMIC-III and MIMIC-IV.

## Test plan
- [ ] Row with NULL `wbc_min`/`bands_max` and present normal `wbc_max` → `wbc_score = 0`
- [ ] All three NULL → `wbc_score` remains NULL
- [ ] Elevated `wbc_max` / low `wbc_min` / high `bands_max` still score 1

### DIFF
--- a/.github/workflows/lint_sqlfluff.yml
+++ b/.github/workflows/lint_sqlfluff.yml
@@ -1,11 +1,16 @@
 name: SQLFluff
 
+# Lints changed mimic-iv/concepts/*.sql on pull requests and posts GitHub
+# annotations. permissions + ignore-unauthorized-error keep fork PRs usable.
 on:
   - pull_request
 
 jobs:
   lint-mimic-iv:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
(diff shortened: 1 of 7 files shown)


alistairewj (approved):
this change is a no-op because wbc_min and wbc_max are always null/not-null together - please try to avoid PRs which "fix" surface level issues that do not actually appear in the data

alistairewj:
needs rebasing

Chessing234:
rebased onto main. the conflict was a sqlfluff workflow commit bundled onto this branch that main has since fixed its own way, so i dropped it — this is now just the one-line coalesce change across the six dialect copies.

PR #2128, opened 2026-07-25
fix: quote load paths, imports, and transpile inputs

Chessing234:
## Summary

Rebased onto current main and squashed to the net change after the earlier rounds of trimming.

- Quote/escape data directories and the created user in the postgres and duckdb loaders (mimic-iii, -ed, -note), so a path or password containing a space or a quote no longer breaks the load
- Fix the mimic-iii concepts build order and the duckdb concepts entry point
- Transpile writes UTF-8 explicitly, with a test
- Correct the table lists in the mysql/postgres `validate_demo.sql`

Scope changes since this was opened, so the old title no longer described the diff:

- The sqlfluff annotate permissions moved to #2148
- The sqlite plain-`.csv` / `.csv.gz` discovery, its unit tests and CI moved to #2149
- The mimic-iv duckdb hunk is dropped: main replaced `mimic-iv/buildmimic/duckdb/import_duckdb.sh` with `build_mimic.sh` in #1757, so that file no longer exists

**This branch touches no GitHub Actions workflow.**

Worth flagging separately: main's replacement `mimic-iv/buildmimic/duckdb/build_mimic.sh:116` interpolates the path straight into `COPY ${table} FROM '${file}'`, so it carries the same unescaped-path problem the dropped hunk fixed. I have not folded that in here — happy to open it on its own if useful.

## Test plan

- [x] `pytest tests/test_transpile.py`

PR #2149, opened 2026-07-31
fix(mimic-iii): load plain CSV in sqlite import.py (prefer .csv.gz)

Chessing234:
## Summary
- Split from #2128 per review: MIMIC-III `import.py` now discovers plain `.csv` as well as `.csv.gz`, preferring gzip when both exist for the same table (parity with `import.sh`).
- Lazy-import pandas so the discovery helpers are unit-testable without a data stack.
- Add `tests/test_mimic_iii_sqlite_import.py` and a PR `unit-tests` workflow so CI verifies the behavior.

## Test plan
- [x] `pytest tests/test_mimic_iii_sqlite_import.py -q`
- [ ] Place demo `.csv` files beside `import.py` and confirm tables load without editing the script
- [ ] When both `ADMISSIONS.csv` and `ADMISSIONS.csv.gz` exist, confirm the gzip path is used

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,19 @@
+name: Unit tests
+
+on:
+  - pull_request
+
+jobs:
+  sqlite-import:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v7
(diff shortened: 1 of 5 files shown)


Chessing234:
@alistairewj sqlite plain-`.csv` / `.csv.gz` discovery split from #2128 with tests/CI as you asked. Ready for a look when convenient.

